Issue #4, opened 2016-10-27
Create an apiClient that integrates with the go-whisk package


paulcastro:
The OpenWhisk API has an existing Go package go-whisk that implements the OpenWhisk REST API.  It is designed to be re-used by other applications.  We should integrate our tool with this package to handle the network calls to an OpenWhisk instance.

go-whisk will provide us with:
- A REST API implementation
- A framework for adding i18n to our code

rabbah:
A good motivation to pull the CLI out of the main repo? @csantanapr FYI

csantanapr:
@paulcastro feel free to work on putting go-whisk on it's own repo, this way you can added as a dependency for wsktool. then wsk cli and wsktool can leverage the same library.

CC @mdeuser

mdeuser:
Some extra go-whisk/go-whisk-cli decoupling may be required to facilitate this work.  I think some utility functions may be shared...

paulcastro:
This has been addressed by new deployer architecture, will close.

PR #6, opened 2016-10-27
WIP for Issue #4, Example API Call for Creating an Action

paulcastro:
This PR provides an example of how to make an API call using the Go whisk package, which provides a full implementation of the OpenWhisk REST API.

In examples/exampleApi.go, we make an API call to create a Swift 3 HelloWorld action.

The flow is:
1. Create a whisk client which is a container for an http client and specific entity operations.  You access these operations using `client.<entityname>.<operation> `.  For example, to create an action we do client.Actions.Insert( .... )
2. Fill out an Action struct that represents the Whisk Action.
3. Call client.Actions.Insert(action, false, false)  where `action` is the Action struct and the booleans represent sharedSet and overwrite, respectively.

Other commands follow this basic model.

### DIFF
--- a/examples/exampleApi.go
+++ b/examples/exampleApi.go
@@ -1,0 +1,66 @@
+package examples
+
+import (
+  "fmt"
+  "net/url"
+  "net/http"
+  "github.com/openwhisk/go-whisk/whisk"
+)
+
+func testCreateAction() {
+  createAction("helloWskTool", "func main(args: [String:Any]) -> [String:Any] { return [\"msg\":\"wsktool says hello\"] }", "swift:3")
+}
+
+func createAction(name string, code string, kind string) {
+
+  baseUrl, err := getURLBase("openwhisk.ng.bluemix.net")
+  if err != nil {
+    fmt.Println("Got error making baseUrl ", err)
+  }
+
+  clientConfig := &whisk.Config{
+      AuthToken:  "MyAuthToken",
+      Namespace:  "MyNameSpace",
+      BaseURL:    baseUrl,
+      Version:    "v1",
+      Insecure:   false, // true if you want to ignore certificate signing
+  }
+
+  // Setup network client
+  client, err := whisk.NewClient(http.DefaultClient, clientConfig)
+  if err != nil {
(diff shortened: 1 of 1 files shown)


examples/exampleApi.go
```diff
@@ -0,0 +1,66 @@
+package examples
+
+import (
+  "fmt"
+  "net/url"
+  "net/http"
+  "github.com/openwhisk/go-whisk/whisk"
```
lzbj: is this package exists, go get complaints not find the repo.

examples/exampleApi.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+func testCreateAction() {
```
lzbj: I suggest we put the test code into separate file, we better obey go best practices

examples/exampleApi.go
```diff
@@ -0,0 +1,66 @@
...
+  createAction("helloWskTool", "func main(args: [String:Any]) -> [String:Any] { return [\"msg\":\"wsktool says hello\"] }", "swift:3")
+}
+
+func createAction(name string, code string, kind string) {
```
lzbj: is it better we put this func to the utiliy package?

examples/exampleApi.go
```diff
@@ -0,0 +1,66 @@
...
+
+  // call ActionService Thru Client
+  _, _, err = client.Actions.Insert(action, false, false)
+  if err != nil {
+    fmt.Println("Got error inserting action ", err)
+  }
+}
+
+// Utility to convert hostname to URL object
+func getURLBase(host string) (*url.URL, error)  {
```
lzbj: same as above.

examples/exampleApi.go
paulcastro: Yes, sorry I copied this manually.

examples/exampleApi.go
paulcastro: This is badly named (by me).  This isn't really a test, just an example of how to use the API.

examples/exampleApi.go
paulcastro: Yes we will need to refactor.

PR #37, opened 2016-11-16
Add deployment file for the 2nd test case and update README.md

daisy-ycguo:


tests/testcases/triggerrule/deployment.yml
```diff
@@ -0,0 +1,14 @@
+application:
+  name: wskdeploy-samples
+  namespace: /wiskdeploy/samples/
```
paulcastro: should we say `/wskdeploy/samples` ?

mrutkows:
Both the manifest and deployment files look good.  However, we recently renamed "location" key under the Action entity to be "function".   This was noted on a recent call, but perhaps went un-noticed.  The goal was to align the name for the functional code artifact (semantics) to other Serverless offerings, but perhaps we should change back to "location" for now that way all entities can have the same keyword.  Let us leave "location" for now and discuss on next specification call with Paul.

mrutkows:
Please note that the project was renamed wskdeploy and your branch is wsktool still

daisy-ycguo:
So I will keep "location" for now, till we decided to change to "function".

PR #40, opened 2016-11-18
#WIP, add local cache for the deployed openwhisk entities.

lzbj:
which can  help user query the deployed entities, then may help to avoid deploy contents repeatedly.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -28,6 +28,8 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	//"runtime"
+	"runtime"
 	"syscall"
 )
 
@@ -37,6 +39,9 @@ var manifestPath string
 var deploymentPath string
 var useInteractive string
 var useDefaults string
+
+// The cached json file.
+var cacheFilePath string
 
 // RootCmd represents the base command when called without any subcommands
 var RootCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


paulcastro:
So the goal here is to serialize the deployment plan as a JSON file, then be able to restore it w/o parsing?  

One concern is that someone may modify their deployments out-of-band from wskdeploy, e.g. use the CLI to delete an action.  This will invalidate the cache.  Perhaps we need a new command to validate or inspect a deployment - we could use the "get" calls in the REST API to download information about the installed actions/trigger/rules and match this to the cache/manifest.

lzbj:
@paulcastro , thank you for your review and constructive comments. Yes, I agree with your concern, I realized perhaps rename it to log instead of local cache maybe more appropriate. So the feature will try to log all the deployments in the past, not rewriting the deployments yaml to json.

The purpose is try to provide a function to help user log the deployments items in the past, perhaps we could use the util function added by you, and change the wskdeploy behavior to let users decide whether to redeploy the duplicated item again instead of just exit if duplication found by default.

I also agree with you we could add "get" to let users get all remote deployments from OpenWhisk REST API and "get" past deployments by local json file, so users could have a rough idea of deployment history even without a openwhisk API connection.

lzbj:
Close it after discussion.

Issue #43, opened 2016-11-23
Support the deployment of trigger & rule


daisy-ycguo:
We need to support the deployment of trigger & rule. 
This function can be tested using tests/testcases/triggerrule

lzbj:
Close as pr merged.

PR #48, opened 2016-11-24
Add deployementService support for rule.

lzbj:
Address issue #43


lzbj:
close duplicated with another pr.

Issue #45, opened 2016-11-24
[WIP]Add basic support of deployment.yml


daisy-ycguo:
We need to start supporting deployment file. According to the Specification, a Deployment file contains configuration information, e.g.
- input parameters
- authorization credentials
- namespaces

and so on.

There is a simple Deployment file in tests/testcases/triggerrule, with namespaces of application and packages, authorization credential of OW, and input parameters to rule. We can start the support of deployment file from this basic case.

lzbj:
Daisy is discussing and confirming with Matt about the sample deployment yaml definition and usage. The code implementation will follow that.

PR #52, opened 2016-12-05
Update deployment yaml files

daisy-ycguo:
Update the namespace settings in deployment.yml
Address issue #45


lzbj:
@paulcastro ,is this pr ok, I'll try to add some code implement as if I can?

paulcastro:
@lzbj sounds good

PR #54, opened 2016-12-06
[WIP] Add support to parse feed

daisy-ycguo:
1. modify yamlparser.go to add supporting of feed
2. modify yamlparser_test.go to add unit test
3. [WIP] update the webhook feed test case


daisy-ycguo:
@mrutkows I think we need to review the manifest.yml and deployment.yml of testcase3. The concept of feed here is different from the "feed" in OW.

PR #63, opened 2016-12-15
Update test case 3

daisy-ycguo:
1. updated webhook.js to the latest code in openwhisk-catalog
2. add a new trigger "webhook-trigger", the trigger is marked as a feed trigger
3. In the deployment.yml file, pass all the parameters to feed.

### DIFF
--- a/tests/testcases/webhook/README.md
+++ b/tests/testcases/webhook/README.md
@@ -1,17 +1,30 @@
 # Test case for WskTool.
 
-This is a test case for WskTool. The package includes:
-- A feed named as "webhook_feed". It will add a webhook to a specified github repository. 
-- A trigger named as "webhook_trigger". When there is some specified actions happened, the trigger will be fired.
+This is a test case for WskDeploy tool. The package manifest file includes:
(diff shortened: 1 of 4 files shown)


lzbj:
@daisy-ycguo , could you please help updates the readme a little, so people could get a better idea about what our test cases doing. we also need to make sure manually our test cases could pass on a openwhisk instance, then we can use the deployment and manifest files to describe it and deploy the packages and entities automatically. 

I think we need to make some clean of our test datas, if we want to support a whole use cases, we better put them under one folder, what do you think?

lzbj:
@daisy-ycguo , I tested this PR with our current code implementation, I found it doesn't work.
I made those changes, and it works.

--- a/tests/testcases/webhook/manifest.yml
+++ b/tests/testcases/webhook/manifest.yml
@@ -1,11 +1,12 @@
-git_webhook:
+package:
+  name: gitwebhook
   version: 1.0
   license: Apache-2.0
   actions:
     webhook:
       version: 1.0
-      function: github/webhook.js
-      runtime: nodejs@6
+      location: src/webhook.js
+      runtime: nodejs
       inputs:
         username:
           type: string

So please use the test cases under the triggerrule as a template before you submit the newer test cases, you can test with this command.

go run main.go -d deploymentfile  -m manifestfile.

daisy-ycguo:
Thank you @lzbj . I updated as your comments.

PR #66, opened 2016-12-20
Implement issue#47 for the feed support.

lzbj:
Temporarily add the feed infos as annotations for the trigger.


lzbj:
Close it as a later pr overlay.

PR #95, opened 2017-02-10
Support existing package

lzbj:
Rss package test cases.


daisy-ycguo:
@lzbj: Thank you for taking care of the rss package use case. It looks like your PR is a composition of several commits.  Is that on purpose or by mistake ? Please check.

mrutkows:
This looks like it picked up some things I already thought were squashed and merged?

lzbj:
Sorry, it's a mistake, I'll close this one and resubmit.

PR #97, opened 2017-02-13
Add test cases manifest and deployment yml file for

lzbj:
RSS package.


paulcastro:
I'll merge, however I'd like to review how feeds work across deployment and manifest files.

lzbj:
Yes, correct, we can add a new issue for this if you think we need reuse feed accross deployment and manifests.

PR #98, opened 2017-02-13
[WIP] Add undeploy cmd, to remove all entities currently deployed

lzbj:
in Openwhisk according to manifest yml.


lzbj:
@paulcastro ,just quick POC, please help review.

I have one question, should we separate the deploy/undeploy to different cmd modules, currently the deploy is under root cmd, and all the deploy yml and manifest yml is under root flag.
If we separate, we should have different flags under diff cmd modules. right?

paulcastro:
Yes, `deploy` and `undeploy` should be separate commands.  We may need different flags for each.

We should also have some verification that compares the manifest to what is deployed.

lzbj:
ok, got it, I'll do it, thank you.

paulcastro:
Apologies, but I did not answer correctly.  `wskdeploy deploy` is strange, so let's keep it in the root command.  However, let's keel `wskdeploy undeploy` as a separate command.

lzbj:
Ok, got it. Well, It doesn't matter.

PR #103, opened 2017-02-15
Add multipackage support, add sequences, refactor

paulcastro:
This PR is for code review, **this should not be merged yet.** It still needs to be debugged and the actual deployment of entities is not completed yet.  Also, it **may be missing some files** since I tore everything down and put it back together.  I'll add back the missing files (like report.go and version.go).  Sorry about that :-)

The goal here is to refactor the code to modularize it better so we can add "big" features.  

The major refactors are:

- refactor utils.go into separate classes
- refactor manifest and deployment functions into parsers and readers
- refactor classes into packages that are more descriptive of their functions, e.g. parsers, utils, deployers

The features added in this code:

- support for multiple packages in the service deployer.  This is mainly targeted towards multiple packages in the deployment.yaml, not the manifest.yaml
- add parameter and annotation binding from the deployment file into the deployment plan.
- add support for sequence notation per use case` openstack.`
- add placeholder for dependency specification in a package.


lzbj:
Wow, this is a huge change, but will definitely improve the wskdeploy quality. Great job.
Just one comment, you don't have to finish all test and features to merge this pr,
since if we refactor the project hierarchy and structures fundamentally, other works are dependent on this. Please open issues to track what's needs to be done, so we could do works parallelly.

paulcastro:
@lzbj yes excellent point.  Give me another day to patch things up, then we can merge and create issue for the missing/new features.

lzbj:
Sure, thank you.

paulcastro:
OK, I've patched it back up.  I have only tested this on the open stack test case.  

@daisy-ycguo, sorry but the web hook use case is likely broken by this PR.  I'll need some help fixing it because I'm unsure what I need to add back to support the feeds.

lzbj:
Thank you @paulcastro , I'll update merge it, let's open new issues to fix the tests later.

Issue #93, opened 2017-02-08
Support 'wskdeploy undeploy'


daisy-ycguo:
The goal would be to read a manifest file, then remove all entities currently deployed in OpenWhisk.

lzbj:
Good feature.

paulcastro:
Do we want to think of having a "verify" feature as part of this (or as a separate issue?).  Verify would read the manifest and deployment files, then check to see what is already installed in OpenWhisk. Verify is probably useful on its own, but I'm thinking that it can be used during deploy or undeploy to update just what has changed.

lzbj:
Good point, is it like a diff tool to help users know the difference between existing entities and entities defined in deployment and manifest? if that's the case, my personnel view is we add it as a separate cmd, to help user determine next step(deploy or undeploy).

lzbj:
@paulcastro , I opened another issue for this, I think it's a step before we implement undeploy.

lzbj:
Close as pr merged.

Issue #107, opened 2017-02-16
[WIP] Add verify command.


lzbj:
Add a verify command which would read the manifest and deployment files, then check to see what is already installed in OpenWhisk. 
This is a derivative of https://github.com/openwhisk/openwhisk-wskdeploy/issues/93

paulcastro:
Sounds good - we can do this in two phases.  The undeploy function can take arrays of entities (from the manifest) that should be undeployed.  In the first phase, we just pass all the entities from the manifest even if they aren't deployed.  In the second phase, we add the verification step which takes the entities from the manifest, and filters out the entities that haven't been deployed and then passes them to the undeploy function(s). 

I can create the undeploy functions if you are already working on verification.

paulcastro:
I made a PR for undeploy #110

lzbj:
Ok, thanks, hope I got a chance to submit a pr regarding this issue. :-)

lzbj:
Close is as I think it's not necessary feature and  on the critical path any more.

PR #110, opened 2017-02-16
Add undeploy

paulcastro:
Fixes issue #93 .  Adds an undeploy option.

The assumption is you should "verify" what is actually deployed (issue #107) and create a DeploymentPackage map that contains a "verified" set that will delete those entries.  We currently don't have verification, so this PR just assumes everything is deployed.  This will generate non-fatal error messages that a whisk.Client command could not succeed because it tried to delete an entity that does not exist.  However, it does succeed in undeploying the existing entities.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,12 +23,17 @@ _testmain.go
 *.test
 *.prof
 
+
+#OS X
+**/.DS_Store
+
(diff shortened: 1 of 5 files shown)


lzbj:
close as merged

Issue #126, opened 2017-02-22
panic: runtime error: invalid memory address or nil pointer dereference


jzaccone:
After following the "How to build on local host" instructions, I ran: 
```sh
./openwhisk-wskdeploy -m tests/testcases/triggerrule/manifest.yml -d tests/testcases/triggerrule/deployment.yml
```

I get the following error:
``` sh
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8a427]

goroutine 1 [running]:
panic(0x3fd0e0, 0xc420012090)
	/usr/local/Cellar/go/1.7.5/libexec/src/runtime/panic.go:500 +0x1a1
github.com/openwhisk/openwhisk-wskdeploy/deployers.(*DeploymentReader).bindPackageInputsAndAnnotations(0xc4201cfbb0)
	/Users/johnzaccone/go/src/github.com/openwhisk/openwhisk-wskdeploy/deployers/deploymentreader.go:69 +0x467
github.com/openwhisk/openwhisk-wskdeploy/deployers.(*DeploymentReader).BindAssets(0xc4201cfbb0, 0x0, 0x0)
	/Users/johnzaccone/go/src/github.com/openwhisk/openwhisk-wskdeploy/deployers/deploymentreader.go:36 +0x2b
github.com/openwhisk/openwhisk-wskdeploy/deployers.(*ServiceDeployer).ConstructDeploymentPlan(0xc4201d2770, 0x1c, 0x7fff5fbffbdd)
	/Users/johnzaccone/go/src/github.com/openwhisk/openwhisk-wskdeploy/deployers/servicedeployer.go:97 +0x7a
github.com/openwhisk/openwhisk-wskdeploy/cmd.glob..func2(0x669d00, 0xc4201a4400, 0x0, 0x4)
	/Users/johnzaccone/go/src/github.com/openwhisk/openwhisk-wskdeploy/cmd/root.go:119 +0x2a7
github.com/spf13/cobra.(*Command).execute(0x669d00, 0xc42000c1f0, 0x4, 0x4, 0x669d00, 0xc42000c1f0)
	/Users/johnzaccone/go/src/github.com/spf13/cobra/command.go:650 +0x443
github.com/spf13/cobra.(*Command).ExecuteC(0x669d00, 0xc4200180b8, 0x0, 0x669d00)
[...]

PR #131, opened 2017-02-23
Fix triggerrule testcase. Removed unused credential field.

jzaccone:
Fixes issue #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,11 @@ $ ./wskdeploy --help
 
 For example,
 ```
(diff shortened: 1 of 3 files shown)


jzaccone:
@paulcastro

paulcastro:
@jzaccone Thanks a lot for the fix!

OW is in an apache incubator now so there might be an additional step before I can merge.@mrutkows does @jzaccone need to send Apache an ICLA for something like this?

PR #133, opened 2017-02-23
Implement filesystemreader for usedefault mode

### LINKED ISSUE #120: Implement filesystem reader as a deployer
The current tool doesn't support file system defaults to deploy OpenWhisk actions.  We can add some conventions to parse a file system directory and add actions to the deployment plan.

We can directly update the plan in memory, or update the manifest file itself and add the actions.  The latter would provide an intermediate staging point to see the entire plan in one place, however it might be disruptive to allow the deploy to change the manifest file.

paulcastro:
Fixes #120 . This implements the option to read files (.js, .py, .swift) from the file system and install them as actions.  We make some changes to how wskdeploy creates a deployment plan:

`wskdeploy --allow-defaults=true` activates this feature.

In this mode, a deployment plan consists of:  1) information from reading a project directory 2) information from the manifest.  The behavior of the deployment file does not change since it is applied to the overall deployment plan.

wskdeploy reads the project directory first
- it looks for source code in the root project directory, and a subdirectory called `actions`
- the filename becomes the action name, the extension determines the kind

wskdeploy reads the manifest file
- any manifest entry for an action is compared to the information from the project directory.  We merge manifest entries that declare actions  that have the same name as a project directory action. Manifest entries always override information from the project directory 
- wskdeploy will throw an error if an action with the same name has two different locations.  For example, action A is derived from reading the file system at location actions/a.js, while the manifest declares action A with location actions/myactions/a.js .

In this mode, it is NOT an error if you do not specify a location for an action in the manifest file as long as the action kind, name, and source code can be derived from the file system.

An example of a minimal set of items you need for this mode is now in testcases/httpget.  I plan to make the deployment file optional -- right now the tool throws an error without it.


lzbj:
close as merged.

Issue #141, opened 2017-02-27
Tool crashes if deployment.yaml is not present


paulcastro:
Tool will crash with nil pointer if deployment.yaml is not present.  This file should not be mandatory since it is primarily used for now to set bindings.  Other needed settings like namespace/credentials can be set in the ~/.wskprops file or entered from interactive prompts.

PR #143, opened 2017-02-27
Make deployment.yaml optional

paulcastro:
This fixes issue #141.  Changes:

Run with default mode on (so it reads file system in addition to manifest.yaml) `wskdeploy --allow-defaults=true `

- allows minimum manifest.yaml to deploy source code (example in use case `defaulttest`)
- allow binding if deployment.yaml is present (example in use case `defaulttest-bindings`)


lzbj:
close as merged.

PR #145, opened 2017-02-28
Just a POC of verifier to cover more complicated use cases.

lzbj:


lzbj:
@paulcastro , please don't merge it, it's just a POC, I considered the verifier for more use cases, and added them in the comments, please help review. Do you think we need to cover all those use cases? Or we make the undeploy simpler? I found currently the wsk client seems not support forcely delete package with entities in it, or the undeploy process will be much more easier?

paulcastro:
Yes you cannot delete an non-empty package.

PR #150, opened 2017-03-02
Add support for feed actions

paulcastro:
Prototype support for curated feed actions by specifying a source for a trigger.

```
triggers:
    myTrigger:
         source: /whisk.system/alarms/alarm
```

Parameters must be set in the deployment.yaml file for this trigger as required by the feed action source (they're all different).

This PR also adds limited support to detect if a string is really JSON for trigger parameters to do the proper conversion.


paulcastro:
I've successfully tested this against the following feed actions:

/whisk.system/alarms/changes
/whisk.system/cloudant/changes
/whisk.system/github/webhook

Issue #146, opened 2017-03-01
Support inline parameter binding in manifest file


paulcastro:
Currently, parameter binding can only be done in a deployment file.  It should be possible to do in-line parameter binding in the manifest, especially for values that are set by the developer and not intended to be changed.

One possible change is to add a value to the schema "inputs" key:

```
actions:
   myAction:
        inputs:
             myParam:
                  type: string
                  value: "Hello World!" 
            anotherParam:
                  type: int
                  value: 1
```

However, we also allow simple name value pairs without types in the deployment file.  We could allow the same in the manifest without using the schema keys:

```
actions:
   myAction:
       parameters:
            myParam: "Hello World"
            anotherParam: 1
```

paulcastro:
My preference is the second option.

mrutkows:
This is supported in the 0.8.3 Packaging Spec. as the first "use case" for hello world: Manifest Files:

**Manifest files**
_Example 1: No parameter descriptions_
```
helloworld:
  version: 1.0
  license: Apache-2.0
  actions:
    hello:
      version: 1.0
      function: src/hello/hello.js
      inputs:
        personName: string
      outputs:
        greeting_string: string
```

paulcastro:
I'd like a simpler approach where the version of the package applies to the entities so we don't need to explicitly assign a version number to actions, triggers, rules.

paulcastro:
Inferred types would be good.  The inputs in the YAML file get converted into a JSON based request body underneath the covers before calling the OpenWhisk REST API so any inference we make explicit in the spec should follow the JSON type inference rules.

paulcastro:
If we overload the inputs/outputs section to include values, I see two options.

**Option 1: Preserve existing keys and add a value key**
```
actions:
    hello:
      version: 1.0
      function: src/hello/hello.js
      inputs:
             name:
                  type: string
                  value: Paul
 ```
**Option 2: Allow name value pairs as inputs**
```actions:
    hello:
      version: 1.0
      function: src/hello/hello.js
      inputs:
             name: Paul
```

Would we allow mixing and matching of the two? Or are they mutually exclusive.  One problem is not all inputs will take parameter values.  If at some point in the future we need to know all the inputs, what is the best way to handle that case?

paulcastro:
merged

PR #151, opened 2017-03-03
Add support for in-line parameters

paulcastro:
Fixes issue #146.

Adds a `parameters` key to the manifest file which allows setting of parameters:

```
actions:
    myAction:
         parameters:
                 name: Paul
                 vehicle: airplane
```

The presence of a deployment.yaml file will overwrite any parameters set in the manifest file.


lzbj:
@paulcastro , there are some conflicts with the master branch after merged your former PR, could you please check it and resubmit?

paulcastro:
@lzbj Resolved

lzbj:
@paulcastro , sorry, still show conflicts when I try to merge it, perhaps you need to rebase the master, resolve the conflicts, then resubmit.

paulcastro:
@lzbj its fine.

mrutkows:
@paulcastro I would prefer not adding "parameters" as a key as it is redundant to inputs; it will break our intended alignment with OpenAPI, as well as long term plans to be part of efforts such as Lambda has in Cloud Formation.

paulcastro:
@mrutkows `Inputs` is still an allowed key in this PR.  Is your change request that we allow value setting in the inputs section? In other words:

```
actions:
    myAction:
        inputs:
                 name: Paul
                 vehicle: airplane
```

Per my comments on #146, it seems cleaner to have a different section to set parameter values than to overload the schema definition using `inputs` and `outputs`.

In the current PR, this is allowed:

```
actions:
    myAction:
         inputs:
                 name:
                       type: string
                 airplane:
                       type: string
         parameters:
                 name: Paul
                 vehicle: airplane
````

paulcastro:
Updated to overload `inputs` as the key to set parameter bindings in the manifest file.  Note that this overloads the schema definition section and will need to be modified if we need to infer schema information from this section in the future.

lzbj:
@paulcastro , yes, merged, I used the github UI to do the merge, then it's succeeded.
If I use the git cli, I always got conflicts errors, Just FYI, I referenced a wiki here:
https://github.com/openwhisk/openwhisk/wiki/Contributing:-Git-guidelines

Issue #160, opened 2017-03-06
Update to use latest openwhisk-client-go


paulcastro:
The openwhisk-client-go needs to be updated to use the latest code from the core.  see [issue 1](https://github.com/openwhisk/openwhisk-client-go/issues/1).  Once that's done, we need to update the wskdeploy code to use it.  There are differences in the latest openwhisk-client-go code:

whisk.Package instead of whisk.SentNoPublish
whisk.KeyVal.Value is now a map[interface{}][interface{}
whisk.Action.Exec.Code is now *string
whisk.Action.Publish is now *bool

There are others.

paulcastro:
Merged

Issue #161, opened 2017-03-06
add i18n to all strings


paulcastro:
When Issue #160 is fixed, we should add i18n to all strings using the wski18n code.  This means the build instructions for wskdeploy will include the packing of the i18n resources (as done by the CLI today).

PR #162, opened 2017-03-06
Fixes issue #160, use latest go-whisk code, restore i18n

paulcastro:
This PR will update the wskdeploy tool to use the latest Go client libraries, and make it possible to do i18n strings (see issue #161 ).

**Important:**  This code will only work with the `i18n` branch of the `github.com/openwhisk/openwhisk-client.go/whisk` repo we're using.  The `i18n` branch contains a snapshot of the latest go-whisk libraries and turns `wsk18in` on again.  Using this branch will restore the proper error messages.

This is a breaking change against `master` of `github.com/openwhisk/openwhisk-client.go/whisk`


paulcastro:
Travis build fails because it loads openwhisk-client-go from master, not i18n branch

Issue #41, opened 2016-11-19
Add support to manage dependencies


paulcastro:
The current tool cannot reference other OpenWhisk packages as dependencies.  At a minimum we should be able to pull in and reference entities in other packages w/o requiring the developer to copy the source code.  This requires:

- create a module to parse dependency graph and validate it
- create feature to download/obtain dependencies into the current project path and be able to reference these from inside the manifest and deployment yaml files.

This is a large task and likely should be broken up into multiple issues.

As a starting step let's propose:
- support for getting a Github-based  OpenWhisk package
- mange a semantic version applied to an entire deployment (as opposed to individual entities)

lzbj:
@paulcastro ,very interesting features, please split it into smaller issues, so we could help if possible.

paulcastro:
When someone pushes a package to a repo, there should be a default package name for it that is unique.  We currently don't have a scheme for this though we could borrow from other dependency management frameworks.  For instance, we could use a URL as the canonical name for the package.

When someone integrates a package from a repo, they should have the option to override the default name.  

An example of support for this kind of model is how Go treats imports.

lionelvillard:
After weighting several options, I would use a shorten (no protocol) URL as the canonical name for the package as you suggested. This has several advantages:
- it is unique
- no need to override the default name to avoid conflict
- no need to go through a global naming service to guarantee the package name uniqueness
- easier server-side deployment
- the name can be used to directly locate the source code 
 
The only minor problem is the package name cannot contains `/` characters, which has a simple solution.

paulcastro:
Thinking about package bindings, perhaps something like this:

```
package:
   name: myPackage
   dependencies:
           Cloudant:
                  version 1.2
                  location: /whisk.system/cloudant
                  binding-name: myCloudant
                  inputs: 
                         param1: value1
 ```

The above assumes there is an OpenWhisk package called `Cloudant`, it's location is "whisk.system", and I want to refer to it as `myCloudant` inside this manifest file.

I'm not sure how to handle name uniqueness for package names.  One option is to follow go's model, and use a internet style name based on whisk.system or GitHub repos:

```
package:
   name: myPackage
   dependencies:
           /whisk.system/cloudant:
                  version 1.2
                  binding-name: myCloudant
                  inputs: 
                         param1: value
          github.com/openwhisk/openwhisk-client-swift
                  version .02.2
                  binding-name: mySwift
 ```

[...]

PR #167, opened 2017-03-07
Add use case for data-processing-app.

lzbj:
Modified the data-processing-app deployment yaml. This is a work still in progress, we need to add the support for the package level bindings.

### DIFF
--- a/tests/usecases/data-processing-cloudant/manifest.yaml
+++ b/tests/usecases/data-processing-cloudant/manifest.yaml
@@ -1,0 +1,34 @@
+package:
+  name: data-processing-cloudant
+  version: 1.0
+  license: Apache-2.0
+  binding-package: /whisk.system/cloudant
+  inputs:
+    username: cloudant_user
+    password: cloudant_password
+    host: cloudant_user.cloudant.com
+  actions:
+    write-to-action:
+      version: 1.0
+      location: actions/write-to-cloudant.js
+      runtime: nodejs
+      inputs:
+        CLOUDANT_USERNAME: cloudant_user # The same as above one
+        CLOUDANT_PASSWORD: cloudant_password # The same as above one
+        CLOUDANT_DATABASE: cloudant_database
+    write-from-action:
+      version: 1.0
+      location: actions/write-from-cloudant.js
+      runtime: nodejs
+  sequences:
+      write-from-cloudant-sequence:
+        actions: /whisk.system/cloudant/read,write-from-cloudant  # cloudant_instance is the same as the package binding name
+  triggers:
+    image-uploaded:
+      feed: /whisk.system/cloudant/changes # cloudant_instance is the same as the package binding name
+      inputs:
+        dbname: cloudant_database # The same as above one
+  rules:
+    echo-images:
+      trigger: image-uploaded
+      action: write-from-cloudant-sequence
--- a/cmd/report.go
+++ b/cmd/report.go
@@ -35,7 +35,7 @@ var wg sync.WaitGroup
 
 // reportCmd represents the report command
 var reportCmd = &cobra.Command{
-	Use: "report",
+	Use:   "report",
 	Short: "Returns summary of what's been deployed on OpenWhisk in specific namespace",
 	Long: `Command helps user get an overall report about what's been deployed
 on OpenWhisk with specific OpenWhisk namespace. By default it will read the wsk property file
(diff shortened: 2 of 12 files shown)


lzbj:
@daisy-ycguo @mrutkows @paulcastro @lionelvillard , I added the latest use case which needs to support the package level bindings, could you please take a review, as we might as needs to update the corresponding specs.

tests/usecases/data-processing-cloudant/manifest.yaml
```diff
@@ -0,0 +1,29 @@
...
+      location: actions/write-to-cloudant.js
+      runtime: nodejs
+    write-from-action:
+      version: 1.0
+      location: actions/write-from-cloudant.js
+      runtime: nodejs
+  triggers:
+    image-uploaded:
+      feed: cloudant_instance/changes # cloudant_instance is the same as the package binding name
+      inputs:
```
paulcastro: we're now overloading inputs to do in-line parameter binding so setting schema information here doesn't have any effect.

lzbj: Yes, updated.

tests/usecases/data-processing-cloudant/manifest.yaml
```diff
@@ -0,0 +1,34 @@
...
+    write-from-action:
+      version: 1.0
+      location: actions/write-from-cloudant.js
+      runtime: nodejs
+  sequences:
+      write-from-cloudant-sequence:
+        actions:  # cloudant_instance is the same as the package binding name
+  triggers:
+    image-uploaded:
+      feed: changes # cloudant_instance is the same as the package binding name
```
paulcastro: The key feed doesn't do anything right now?  I implemented the feed action API which uses the key `source` and takes the fully qualified name as the value, in this case it would be `/whisk.system/cloudant/changes`.

I like the key `feed` better than `source.`. Should I use this for feed actions?

tests/usecases/data-processing-cloudant/manifest.yaml
```diff
@@ -0,0 +1,34 @@
+package:
+  name: data-processing-cloudant
+  version: 1.0
+  license: Apache-2.0
+  binding-package: /whisk.system/cloudant
```
paulcastro: I assume this means the entire package is bound to this name.  This would prevent us from adding custom entities to this package

paulcastro:
@lzbj  this is a nice start.  I have a few questions.

If I understand correctly, this PR will allow a package to be bound to another package using the current packages name.  Wouldn't this disallow custom code? 

Currently, if you bind to a package, you are not allowed to modify that package.  e.g.

1. `wsk package bind myCloudant /whisk.system/cloudant`
2. `wsk action create myCloudant/myAction actionCode.swift`

Step 2 will yield` Operation not permitted on package binding`

Am I correct?  If so, this PR will only cover the use case where you just want to use a shared package and not add any of your own entities.

I wonder if package binding should be folded into #41 , where we add dependencies.  In this scenario, the binding adds a reference to installed code in OW, and then the binding name is an alias for the actual package name.

lionelvillard:
Why would you want to extend an existing package? What's the issue with adding custom code to another package?

paulcastro:
Binding is a naming alias scheme.  If we bind a package using this technique, you cannot extend it (limitation of the current API).

So if you want to build a package that uses a binding to link into another package (like Cloudant), this use case is not supported here.

lionelvillard:
In general custom alias is needed because the same package can be instantiated (or bound) multiple times with different parameters. 

However I wonder if there is a need for extending an existing package. From a source code management point of view, it's better to keep the original package content separated from the extensions.

paulcastro:
I added some comments about how we might implement this in a dependencies section in #41.

tests/usecases/data-processing-cloudant/manifest.yaml
lzbj: Yes, agree with you.

lzbj:
@paulcastro @lionelvillard ,thanks for your comments. Yes, I just want to support the cloudant use case, I'll send you the link to the shell script in our slack as put here is not appropriate,  

Here is as my understanding of the use case:
The use case just wants to reuse some existing /whisk.system/cloudant action, with some custom actions, then create an action sequence mixed of existing cloudant action with custom action, then create a trigger with existing action as a feed, finally create a rule to associate trigger and action sequence. So the custom action might not necessarily bind to a custom pacakge, right?
But yes, I agree. Anyway, if we want to better support versatile use cases and potential combinations, we need to think of a better yaml model to describe dependencies and then support them.

I updated the yaml file, with the log after I deployed with the latest code.
[log.txt](https://github.com/openwhisk/openwhisk-wskdeploy/files/837207/log.txt)

PR #176, opened 2017-03-11
Add Publish command. 

lionelvillard:
For now it only registers the package to a configurable OpenWhisk Hub. The registry URL is stored in ~/.wskprops.

### DIFF
--- a/cmd/publish.go
+++ b/cmd/publish.go
@@ -79,8 +79,12 @@ var publishCmd = &cobra.Command{
 			owner := paths[l-2]
 
 			// Send HTTP request
-			_, err := http.Get(registry + "?owner=" + owner + "&repo=" + repo)
+			client := &http.Client{}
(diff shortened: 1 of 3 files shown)


cmd/publish.go
```diff
@@ -0,0 +1,94 @@
...
+			if l < 2 {
+				log.Print("Fatal error: malformed repository URL in manifest file :" + repoURL)
+				return
+			}
+
+			repo := paths[l-1]
+			owner := paths[l-2]
+
+			// Send HTTP request
+			_, err := http.Get(registry + "?owner=" + owner + "&repo=" + repo)
```
csantanapr: Why a get instead of post?

csantanapr: Or put

lionelvillard: That's an oversight regarding OW web actions. It should be a PUT (POST is not idempotent).

PR #190, opened 2017-03-16
Adding test case for alarm trigger

pritidesai:
Adding a sample manifest and deployment to demonstrate how to use wskdeploy to create a alarm feed trigger.


paulcastro:
@pritidesai rebase with master?

PR #199, opened 2017-03-20
Enhancement of the travis ci with integration tests

houshengbo:
This patch brings in the following changes:
* Classify the test cases into unit and integration tests with the tags
  unit and integration. In order to run different type of test, we can
  differentiate them by adding -tags=unit or -tags=integration in the
  end of the command. For example, if we want to run the integration tests
  only, we issue the command: go test ./... -tags=integration.
* Add the README under the tests folder.
* This patch adds only one integration test for the trigger rule. More integration tests can be added in future. As long as they follow the same instruction as this one, they will run automatically in Travis CI.


.travis.yml
```diff
@@ -6,12 +6,17 @@ go:
   - 1.7
 
 install:
+  - export BUILD_READY=false
```
cclauss: Would it make sense to use Travis [`env:`](https://docs.travis-ci.com/user/environment-variables/#Defining-public-variables-in-.travis.yml) to define the BUILD_READY environment variable?

mrutkows: @cclauss I would really want to address these as other issue(s); this PR has been open tool long IMO.  The Python 3 change should have been handled as separate PR as well.  We cannot keep prolonging a working/building PR with feature/scope creep.

houshengbo:
@lzbj I have setup the backbone of travis ci to deploy openwhisk, and run the integration tests with
"make integration_test", since I run the test cases via tags. If we tag the test cases with integration at as the top line, we are able to run them with the "make integration_test" or go test ./.. -tags=integration. 

Next, we need to work on a real integration test for wskdeploy. Please continue to work on integration_test.go under tests/src/integration if possible to see what can be added there.

I am thinking of writing a test, calling the code via the manifest and deployment files, fetching the results to see if the action or package is created in openwhisk, and deleting the installed package and action. I know we are using cobra package to orchestrate the commands, but so quite sure what works best so far. Will dig into it tomorrow. If you can take a look at it today, that will be great.

Makefile
```diff
@@ -21,12 +21,17 @@ updatedeps:
...
 	@echo "Testing"
-	go test ./...
+	go test ./... -tags=unit
 
 # Build the project
 build: test
 	go build ${LDFLAGS} -o ${BINARY}
 
+# Run the integration test against OpenWhisk
+integration_test:
```
lzbj: is integration_test depends on unit test better?

install_dependencies.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
lzbj: is this script duplicated?

.travis.yml
```diff
@@ -1,16 +1,36 @@
...
+
 install:
+  - export DEPLOY_BUILD_READY=false
   - go get -u github.com/golang/lint/golint
+  - go get -u github.com/stretchr/testify
 
 script:
   - make lint
   - make build
+  - source ./install_dependencies.sh
```
lzbj: do we need a test step to make sure the dependencies are ready? for example a curl command to test the openwhisk api is ready, or this step is not necessary?

lzbj:
@houshengbo , very appreciate for help add the CI, yes, I reviewed the PR with my comments inline. Just FRY.

.travis.yml
houshengbo: I am thinking of it as well about how to check the openwhisk is ready. This can come later. If openwhisk is not ready, integration tests will fail anyway.

mrutkows:
@houshengbo thanks Vincent!

houshengbo:
The travis build fails at this moment. Here is the reason:

I tried to install openwhisk in travis, and launch the integration tests I wrote. However, wskprops file is missing, and I am missing auth, apihost, apiversion. Do you know how to configure them for wskdeploy? There is one key for whisk.system, and one for guest under openwhisk_home, do we need to pick up one under openwhisk_home? Or do we have other ways to configure it?

@paulcastro @lzbj

lzbj:
@houshengbo , thanks for your updates. Basically what I understand is: your pr added new wrappers for the current command such as root,deploy,undeploy etc, in order to reuse them, right?

If we only needs to support CI, I suggest you make 1-2 prs instead of just one pr to get everything up, first is to make sure the openwhisk is ready. Another one perhaps is add some shim/wrapper to support CI as you did, which si arguable could be implemented in different ways. Just FRY. @paulcastro @mrutkows @daisy-ycguo , any comments?

houshengbo:
I finally make the integration test run in Travis since the namespace, base url and credential ca be configured in deployment.yml. 

I can split the PR later for easy review. 

@lzbj Do you know how to check whether the action, package are available in openwhisk after running wskdeploy in the integration test I wrote?

lzbj:
@houshengbo , you could check if there is any error after the deploy command. No error means success. Do we need tear down? So if every CI is fresh new, we don't need to bother with that part right?

houshengbo:
@lzbj The condition is that we 100% believe in the wskdeploy command. When it shows no error, we can 100% be sure all the entities are deployed in openwhisk.

I am just skeptical whether we should make this assumption that wskdeploy is always right.

lzbj:
@houshengbo , we used wsk cli, it report errors back to wskdeploy.

houshengbo:
@lzbj OK. We can consider wskdeploy equivalent to CLI in terms of precision. The only thing we need to be sure of is catch all the message in CLI.

Issue #189, opened 2017-03-16
Allow `wskdeploy` to run as a OpenWhisk action


lionelvillard:
`wskdeploy` should detect it runs within a `OpenWhisk` action and treat the first argument as a dictionary describing the command to run. It should also return a dictionary.

When running in `OpenWhisk`, `wskdeploy` should not rely on the filesystem to get the OW properties (`~/.wskprops`), the manifest and deployment files. The main arguments should instead be used.

Some commands cannot be run in an action, i.e `init` and `add`. 

This can be used as the basis for #129.

lionelvillard:
Forgot to mention this issue in commit msg: 32e1acf82587ee470e0f44c20aa9d0b1d13ab80c

PR #200, opened 2017-03-20
wskdeploy can now run as an openwhisk action.

lionelvillard:
This PR enables wskdeploy to run as an OpenWhisk action inside a go docker container. (#189)

- New flags have been added:

      --apihost HOST         whisk API HOST
      --apiversion VERSION   whisk API VERSION
      -u, --auth KEY             authorization KEY

- Modified flags:
      -m now accept URL pointing to the manifest.

- Read action content either from local file system or remote http location.
- Fix duplicated persistent flag declarations.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -31,7 +31,11 @@ import (
 
 	"regexp"
 
+	"encoding/json"
+	"errors"
 	"github.com/openwhisk/openwhisk-client-go/whisk"
+	"github.com/openwhisk/openwhisk-client-go/wski18n"
+	"strings"
 )
 
 var RootCmd = &cobra.Command{
@@ -69,7 +73,7 @@ wskdeploy without any commands or flags deploys openwhisk package in the current
 
 		}
 
-		if utils.FileExists(manifestPath) {
+		if utils.MayExists(manifestPath) {
 
 			var deployer = deployers.NewServiceDeployer()
 			deployer.ProjectPath = projectPath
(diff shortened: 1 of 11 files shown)


mrutkows:
@lionelvillard very cool.  Do you have a practical example that you would care to promote?

mrutkows:
Would suggest changing/adding -m to --manifest (short single letter switches are usually reserved for toggle functions, filenames usually are part of the syntax or spelled out for files).

mrutkows:
Also, we would need to support an optional --deployment <file>

lionelvillard:
-m and -d are existing flags. I suggest to open separate issue to consider changing the wskdeploy syntax.

As for the practical example, what about the simple helloworld usecase? @mrutkows Do you have something else in mind

mrutkows:
@lionelvillard just because something is cool doesn't mean we should do it unless we have a customer (or end user) use case to show a practical example.

lionelvillard:
You are right, we haven't described the practical use cases somewhere. I'm currently working on the `deploy application on Bluemix` feature accessible from the OpenWhisk Hub. There are also several practical use cases, like automatic deployment after github release, server-side deployment of large assets, travis integration, etc..

jthomas:
I haven't been involved with this project so I'm sure I should be doing the review 😎

This does sound like a good idea to me, will help to build serverless deployment pipelines.

paulcastro:
@lionelvillard LGTM, one small comment.  Should we be more consistent with using `log` vs `fmt`?  When we're in interactive mode, mixing logs and fmts sometimes causes bad printing in the  terminal, e.g. error messages get printed over prompts, etc.  What do you think?

lionelvillard:
@paulcastro It seems that in Go neither os.Stdout nor os.Stderr is buffered. So mixings stderr/stdout should not be an issue.

paulcastro:
Ok great - I had trouble before and had to switch some statements from log to fmt to get things to be pretty.  I haven't tried what happens if I dump fmt completely.  Just want things to look nice for the user :-)

Issue #205, opened 2017-03-26
Unable to parse baseURL in the application section of deployment.yml


houshengbo:
This is the a valid yml file we ought to parse:

```
application:
  name: wskdeploy-samples
  namespace: guest
  credential: <key>
  baseUrl: https://172.17.0.1/api
```
However, baseUrl value is unable to be picked up by the parser. Values from namespace and credential can be parsed, but the value of baseUrl is nil, though the baseUrl is not empty.
@lzbj @paulcastro This should be a valid bug.

PR #210, opened 2017-03-27
Add the missing double quote to the field baseUrl for deployment application

houshengbo:
Since there is a missing double quote for the field baseUrl, it is unable
to be parsed. This PR adds the quote to field baseUrl making it able to
be parsed.

Besides, this PR adds the test coverage to parse baseUrl and credential.
Closes-Bug: #205

### DIFF
--- a/tests/dat/deploy1.yaml
+++ b/tests/dat/deploy1.yaml
@@ -1,4 +1,6 @@
 application:
   name: wskdeploy-samples
   namespace: /wskdeploy/samples/
+  credential: user-credential
+  baseUrl: https://172.17.0.1/api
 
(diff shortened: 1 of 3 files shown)


tests/dat/deploy1.yaml
```diff
@@ -1,4 +1,6 @@
 application:
   name: wskdeploy-samples
   namespace: /wskdeploy/samples/
+  credential: 23bc46b1-71f6-4ed5-8c54-816aa4f8c502:123zO3xZCLrMN6v2BKK1dXYFpXlPkccOFqm12CdAsMgRU4VrNZ9lyGVCGuMDGIwP
```
paulcastro: I assume this is not a read credential?

houshengbo: Oh, I nee to change it to something less sensitive.

csantanapr: @paulcastro that key is not sensitive, it's the default key for user `guest` from openwhisk core repo used for testing.
https://github.com/openwhisk/openwhisk/blob/master/ansible/files/auth.guest

Issue #172, opened 2017-03-09
In testcases, manifest files with input/output value 'string' treat it as default value, not type specification


mlinnem:
In manifest files, when specifying input and output parameters, the value 'string' was expected to be treated as a type specification, but it is instead treated as a default value.

```
/[manifest.yaml/]
helloworld:
      version: 1.0
      location: src/greeting2.js
      runtime: nodejs:6
      inputs:
        name: string
        place: string
      outputs:
        payload: string

```
```
../../../wskdeploy -m manifest.yaml
wsk action invoke helloworld/helloworld -p name Micah -b
```

Expected Result:
`"Hello, Micah from !"
`
Actual Result:

`"Hello, Micah from string!"`

Either this is a bug, or else it might be better make the value something more like "some_string" or "default_value" or something along those lines, to clarify the purpose.

paulcastro:
see response in #173

mlinnem:
Seems like this might be more an issue for the default values in the serverless-apis sample, rather than a spec problem. @krook thoughts? Again, not for Interconnect necessarily, just a possible improvement.

paulcastro:
This is not really a bug right?  We decided to use `inputs` for parameter binding, not schema definition per discussion in #146

lzbj:
Close as not a bug.

PR #215, opened 2017-03-28
Fix bug #172.

lzbj:


paulcastro:
@lzbj I think issue #216 and issue #219 may supercede this.  Sorry for the confusion.

lzbj:
@paulcastro, no worry, it's ok, thanks for clarification and review.

Issue #219, opened 2017-03-28
Add support for non-inlined parameters


lionelvillard:
This is option 2 of #216:

option 2 (object)
```
<parameterName>:
  type: <string>
  value: <any>
  description: <string>
  required: <boolean>
  default: <any>
  status: <string>
  schema: <OpenAPI Schema Object>
```
This is needed for non-string values

PR #221, opened 2017-03-29
Treat inline parameter value as JSON and add support for expanded par…

lionelvillard:
…ameters (#219)


houshengbo:
@lionelvillard 
FAIL	github.com/openwhisk/openwhisk-wskdeploy/tests/src/deployers [build failed]
There is test case failing.

Issue #232, opened 2017-04-07
Issues when deploying a nodejs package directory with node_modules dependencies 


cjelger:
With the latest changes (https://github.com/openwhisk/openwhisk-wskdeploy/commit/34bcfc6b2d46fe055b018a3931371b13682b3d1f) from @lzbj it's now possible to deploy a directory for an action. This is great but I've seen 2 issues:
- the deployment works, but if the nodejs action contains a node_modules directory for dependencies (even if they are not used in the main code), it's not possible to run the action. It just blocks and no error is displayed. Removing the node_modules sub-directory solves the issue.
- if the node_modules directory contains symbolic links, the deployment fails. Symlinks are very useful to support cross-module dependencies inside a nodejs project, so it would be great to support that.

Thx in advance

cc: @paulcastro as discussed on slack

lzbj:
@cjelger yes, I'll investigate and fix it if I can. thanks

paulcastro:
Merged

PR #235, opened 2017-04-11
Zip action whole folder instead of just files.

lzbj:
Address issue #232


paulcastro:
@lzbj LGTM

Just for clarification, does zipping the whole directory address symbolic link and node_modules issues in #232?

lzbj:
Yes, symbolic links also be zipped.

PR #236, opened 2017-04-11
Add GitHub based dependencies key to deploy and undeploy

paulcastro:
This is a first cut at adding dependencies to a manifest.yml file.  This adds a `dependencies` key where the dependency is a GitHub repo.

```
package:
  name: opentest
  dependencies:
      hellowhisk:
          url: https://github.com/paulcastro/hellowhisk
          version: 1.0.1
      myCloudant:
           source: /whisk.system/cloudant
           inputs:
                  dbname: MyGreatDB
  sequences:
    mySequence:
      actions: hellowhisk/greeting, hellowhisk/httpGet
  triggers:
    myTrigger:
  rules:
    myRule:
      trigger: myTrigger
```
This manifest references a GitHub project aliased as "hellowhisk", version 1.0.1 at the given URL.  If `version` is not specified, it will pull from `master`.

Dependencies that specify a `source` are interpreted as bindings, and we do a package bind.  `url` specifies a GitHub dependency and is treated as an independent deployment.  For example, the root package `opentest` refers to entities in the `hellowhisk` package using the package name format.  The following happens on deploy

1. wskdeploy downloads and unpacks a zip file of the gihub repo in `$ProjectPath/Packages/<dependencyname>`
2. Deploys dependencies first
3. Deploys root package

This PR does not include dependency graph management so use at your own risk.

There is a use case in tests/usescases/deptest that illustrates a project that has no local source code, just a manifest that combines the entities in a dependency.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -22,7 +22,7 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/openwhisk/openwhisk-wskdeploy/cmdImp"
+	"github.com/openwhisk/openwhisk-wskdeploy/deployers"
 	"github.com/openwhisk/openwhisk-wskdeploy/parsers"
 	"github.com/openwhisk/openwhisk-wskdeploy/utils"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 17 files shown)


lzbj:
@paulcastro , could you pls add some comments to the main functions, and add/alter CI test code to make it pass? tks.

paulcastro:
Superceded by #239 from @houshengbo

mrutkows:
@paulcastro Did we ever talk about supporting this in the spec.?

Issue #226, opened 2017-03-31
Add Support for Package Binding


pritidesai:
We have many deployment use cases which uses package binding, for example, the most common package binding is cloudant. Whisk deploy does not have support for such package binding. I am creating this issue to keep track of adding this support. 

Openwhisk CLI  has support for such binding at https://github.com/openwhisk/openwhisk-client-go/blob/master/whisk/package.go#L54

Details on package binding is available at https://console.ng.bluemix.net/docs/openwhisk/openwhisk_packages.html#openwhisk_packages

@paulcastro @mrutkows @daisy-ycguo any comments or feedback are welcome!

mrutkows:
@pritidesai We have discussed adding grammar for "built-in" feeds/actions and adding credentials for alreaedy running (external services), but have not gotten around to formalizing the discussion.  Cloudant, for example, is just one Cloud storage (API) service and we need a general way to integrate with them.  Please note the wskdeploy tool's scope is limited to OpenWhisk APIs and any other service we cannot add deployment support for (and not try to become an "omni" deployer; that is not our role).  However, we can better integrate running/existing services. Thoughts?

lionelvillard:
I'm quite confused. @pritidesai are you referring to #158?

pritidesai:
It has a reference to #41, I am not sure if #41 is about creating package binding listed under "dependencies", please clarify.

I am referring to adding support in wskdeploy so that we can automate these wsk CLI:

wsk package bind /whisk.system/cloudant TrackPRsInCloudant --param-file cloudant.json

This cloudant package can than be used in multiple custom actions. 

wsk package bind /whisk.system/slack PostPRToSlack --param-file slack.json

Reference:
https://github.com/openwhisk/openwhisk-GitHubSlackBot#step-2-create-cloudant-package-binding---trackprsincloudant

mrutkows:
Priti indicates this is resolved.

PR #237, opened 2017-04-13
Modify the change back to use cmdImp

houshengbo:
Fixes issues #41, #226, add github and binding dependencies key to deploy and undeploy


paulcastro:
@houshengbo does this PR supercede #236?  I can close if so.

Seems like tests here still fail?

Issue #233, opened 2017-04-07
Add support for web actions


cjelger:
It would be great if wskdeploy would support web actions (like wsk does).

cc: @paulcastro as discussed on slack

lzbj:
ok, I'll investigate this one.

paulcastro:
@lzbj we have code now that can read annotations in a deployment file.  We should probably copy this code into the manifest reader and refactor the deployment file code to only add new annotations or overwrite existing ones from the manifest.

After that, one way to support web-export is simply make it an annotation:


```
actions:
    myAction:
         annotations:
                 web-export: true
```

@starpit had some thoughts about this model.  While we use annotations in the REST API to set web-export, we might want to support a higher level key in the manifest instead of exposing annotations, for example one way (maybe not a great way):

```
actions:
      myAction:
           web-export: true
```

cjelger:
By the way, the first proposal from @paulcastro is exactly how they do in serverless, see https://serverless.com/framework/docs/providers/openwhisk/guide/web-actions/

lzbj:
Just my personal view, the second one is more intuitive and reasonable.

mrutkows:
Thanks David. Indeed, mainstreamed features should not be handled via annotations (although this may be the case) and reflecting them in the YAML as an attribute (key) of an Action would also be my preference.

mrutkows:
@cjelger support also added to the 0.8.4 spec.  https://github.com/openwhisk/openwhisk-wskdeploy/blob/master/specification/openwhisk_v0.8.4.pdf

lzbj:
close as merged.

cjelger:
Thank you all for that very useful feature!

PR #242, opened 2017-04-18
Add web action support with simple format.

lzbj:
Address issue #233.

### DIFF
--- a/parsers/manifest_parser.go
+++ b/parsers/manifest_parser.go
@@ -236,8 +236,11 @@ func (dm *YAMLParser) ComposeActions(mani *ManifestYAML, manipath string) (ar []
 			keyValArr = append(keyValArr, keyVal)
 		}
 
-		if len(keyValArr) > 0 {
-			wskaction.Annotations = keyValArr
+		// only set the webaction when the annotations are not empty.
+		if len(keyValArr) > 0 && action.Webexport == "true" {
+			//wskaction.Annotations = keyValArr
+			wskaction.Annotations, err = utils.WebAction("yes", keyValArr, action.Name, false)
+			utils.Check(err)
 		}
 
 		wskaction.Name = key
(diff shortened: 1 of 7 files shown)


parsers/manifest_parser.go
```diff
@@ -222,8 +222,11 @@ func (dm *YAMLParser) ComposeActions(mani *ManifestYAML, manipath string) (ar []
 			keyValArr = append(keyValArr, keyVal)
 		}
 
-		if len(keyValArr) > 0 {
-			wskaction.Annotations = keyValArr
+		// only set the webaction when the annotations are not empty.
+		if len(keyValArr) > 0 && action.Webexport == "true" {
+			//wskaction.Annotations = keyValArr
+			wskaction.Annotations, err = utils.WebAction("yes", keyValArr, action.Name, false)
```
paulcastro: This might get overwritten if someone sets an annotation in the deployment file.

tests/src/integration/webaction/webaction_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"github.com/openwhisk/openwhisk-wskdeploy/cmdImp"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+)
+
+// TODO: write the integration against openwhisk
+func TestWebAction(t *testing.T) {
```
houshengbo: Please be advised that we have changed the integration tests to run against binaries.

parsers/manifest_parser.go
lzbj: So if we use option 2, we don't need to consider this, right?

tests/src/integration/webaction/webaction_test.go
lzbj: Yes, updated as the reference.

daisy-ycguo:
merged and close.

Issue #101, opened 2017-02-14
Support "composition entity"


daisy-ycguo:
Use case "[Data Processing with MessageHub](https://github.com/IBM/ogs-data-processing-messagehub)" needs an action sequence. We should support `composition entity` in wskdeploy.

Sample:
```
my_action_sequence:
    type: sequence
    sequence:
        - action_1
        - action_2
        - action_3
    inputs:
        simple_input_string: string
    outputs:
        annotated_output_string: string
```

paulcastro:
Could you take a look at the openstack use case?  There is an example there for sequences also of the form:

```
 sequences:
    authorizedBackup: 
      actions: getApiToken, getServers, createBackup, callSlack
```

I wanted to have a separate entity for sequences since we may want to have more composition entities in the future.  What do you think?

I'm fine with having a formal YAML list or a comma delimited list to specify the actions that make up the sequence.

daisy-ycguo:
I'm fine with your sample. I think yours looks simpler than the previous one which I copied from the pkg specification file.

@mrutkows do you agree to use a separate entity for sequence? If you agree, the specification file needs to be changed accordingly.

paulcastro:
Sequence support is merged in PR #103

daisy-ycguo:
Close this issue. Open another one to reflect the changes to spec.

PR #247, opened 2017-04-24
Fix #246

### LINKED ISSUE #246: Sequence annotations are ignored
Manifest causing the error:
```
package:
  name: test-public 
  sequences:
    protected:
      actions: test-public/web,test-private/sensitive
      annotations:
         web-export:true
  ...
```

I'm getting this error
```
Got error creating action with error message: Sequence component does not exist. (code 185541) and error code: 144.
```

lionelvillard:


mrutkows:
@lionelvillard @paulcastro The use of the top-level key "sequences" has not been discussed at the spec-level and we should not be adopting this as canon into code.  As we have seen there are other "compositions" coming soon and it would be bad precedent to give each top-level recognition within the manifest (by name).   

From a schema point of view, looking at compositions of compositions (using imports/references), we would want shared schema and single indicator that we are encountering grammar that requires "wiring" and additional validation (type checking) and perhaps implicit logic functions (or sequence processing actions/feeds/triggers) to be "filled in" by the deployer (initially, explicitly) and perhaps at the server (implicitly).

paulcastro:
@mrutkows this was discussed in issue #101, which may have predated the `specification` tag.

mrutkows:
@paulcastro we already have a generic grammar for compositions that should have been used first

PR #268, opened 2017-05-11
Update Dockerfile

pritidesai:
Fixing repo name on while restoring Go dependencies, otherwise fails with an error "`can't cd to /src/github.com/apache/incubator-wskdeploy`" while running `./gradlew distDocker`.


lzbj:
thank you priti, could you please rebase?

PR #278, opened 2017-06-02
Fixing JAR failure

### LINKED ISSUE #276: Action creation with .jar fails with "The request content was malformed"
**1: Create a Java file called `Hello.Java` from [here](https://console.ng.bluemix.net/docs/openwhisk/openwhisk_actions.html#openwhisk_actions_java)**

```
import com.google.gson.JsonObject;
public class Hello {
    public static JsonObject main(JsonObject args) {
        String name = "stranger";
        if (args.has("name"))
            name = args.getAsJsonPrimitive("name").getAsString();
        JsonObject response = new JsonObject();
        response.addProperty("greeting", "Hello " + name + "!");
        return response;
    }
}
```

**2: Compile `Hello.java` into a JAR file `hello.jar` as follows:**

```
javac Hello.java
jar cvf hello.jar Hello.class
```

**3: Create `manifest.yaml` with:**

```
package:
    name: helloworldjar
    actions:
        helloworld:
            location: src/hello.jar
            runtime: java
            main: Hello
```

**4: Deployment fails with:**

```
Response body received:
{
  "error": "The request content was malformed:\n'main' must be a string defined in 'exec' for 'java' actions",
  "code": 4001042
}
2017/06/01 12:48:50 Got error creating action with error message: The request content was malformed:
'main' must be a string defined in 'exec' for 'java' actions (code 4001042) and error code: 144.
```

pritidesai:
Adding support for command line argument "--main" which is used to specify the name of the action entry point function when such function is part of a class in case of Java runtime.

Closes #276


parsers/manifest_parser.go
```diff
@@ -274,6 +274,9 @@ func (dm *YAMLParser) ComposeActions(mani *ManifestYAML, manipath string) (ar []
 			wskaction.Exec.Kind = action.Runtime
 		}
 
+                // we can specify the name of the action entry point using main
+                wskaction.Exec.Main = action.Main
```
lzbj: what if the action does not have a main function, do we need a if test here, what do you think?

mrutkows: we should test that we can support Lambda "handlers", that is reassigning "main"

exports.main = “myAWSLambdaHandler”

which is allowed...

lzbj: @pritidesai , could you pls answer to matt's comments?

pritidesai: Here the main function is missing from the action file and therefore we are using "--main" parameter of `wsk` CLI to designate other function as main. Also, not sure how we can produce an error message from wskdeploy when main function is missing from the action file and `--main` parameter is not populated. `wsk` CLI has this information and does error out saying "'main' must be a string defined in 'exec' for 'java' actions".

Let me know if I am missing anything. Happy to fix it.

Issue #297, opened 2017-07-20
WIP: supplying "type" key-value in Action in manifest causes Nil pointer exception


mrutkows:
Run the following wskdeploy use case, using this command:
```
go run main.go -m tests/usecases/alarmtrigger/manifest.yaml
```

you get this stacktrace:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x100 pc=0x133e51e]
goroutine 1 [running]:
github.com/apache/incubator-openwhisk-wskdeploy/utils.GetEnvVar(0x0, 0x0, 0x100abc3, 0xc4200f9538)
	/Users/Matt/go/src/github.com/apache/incubator-openwhisk-wskdeploy/utils/misc.go:136 +0x5e
github.com/apache/incubator-openwhisk-wskdeploy/parsers.ResolveParameter(0xc4200f95f8, 0xc42019c990, 0xc4200f9538)
	/Users/Matt/go/src/github.com/apache/incubator-openwhisk-wskdeploy/parsers/manifest_parser.go:408 +0x3c
github.com/apache/incubator-openwhisk-wskdeploy/parsers.(*YAMLParser).ComposeActions(0xc420125480, 0xc4201ca000, 0x7fff5fbffb53, 0x29, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/Matt/go/src/github.com/apache/incubator-openwhisk-wskdeploy/parsers/manifest_parser.go:287 +0x813
github.com/apache/incubator-openwhisk-wskdeploy/deployers.(*ManifestReader).HandleYaml(0xc4200f9980, 0xc4201583f0, 0xc420125480, 0xc4201ca000, 0x0, 0x0)
	/Users/Matt/go/src/github.com/apache/incubator-openwhisk-wskdeploy/deployers/manifestreader.go:65 +0xb2
github.com/apache/incubator-openwhisk-wskdeploy/deployers.(*ServiceDeployer).ConstructDeploymentPlan(0xc4201583f0, 0x15, 0xc42012e9c0)
	/Users/Matt/go/src/github.com/apache/incubator-openwhisk-wskdeploy/deployers/servicedeployer.go:131 +0x116
[...]

PR #298, opened 2017-07-20
Fix Parameter Resolution (both schema) & documentation for the triggerrule use case

mrutkows:


mrutkows:
THis PR may also fix
 https://github.com/apache/incubator-openwhisk-wskdeploy/issues/297

mrutkows:
@houshengbo @pritidesai @daisy-ycguo please review and merge.  Also, @priti can you please take the 2 manifests I created here (i.e., tests/dat/manifest_validate_multiline_params.yaml and tests/dat/manifest_validate_singleline_params.yaml) and turn these into unit test cases for the manifest_parser.go file (for unit testing the ResolveParameter() function)?

pritidesai:
thanks @mrutkows yup have been working on creating those unit tests, will add them here in this PR.

pritidesai:
Matt, I have tested your changes and have uploaded test cases here (marked it as WIP as adding more test cases):

https://github.com/apache/incubator-openwhisk-wskdeploy/pull/308

parsers/manifest_parser.go
```diff
@@ -403,19 +425,151 @@ func (action *Action) ComposeWskAction(manipath string) (*whisk.Action, error) {
...
+		// we have a multi-line parameter declaration
+
+		// if we do not have a value, but have a default, use it for the value
+                if param.Value == nil && param.Default !=nil {
+			param.Value = param.Default
+		}
+
+		// if we also have a type at this point, verify value (and/or default) matches type, if not error
+		// Note: if either the value or default is in conflict with the type then this is an error
+		tempType, errorParser = ResolveParamTypeFromValue(param.Value)
```
pritidesai: Here the comment says error should be thrown if we have a mismatch of specified "type" and "type" of the value. Try running it with an example of Parameter{Value: 11, Type: string}. ResolveParamTypeFromValue(11) determines the normalized type name which is "integer" and sets tempType to "integer". There is no check on actualType and type of the value done. This is how struct Parameter looks like:

```
(*parsers.Parameter)(0xc420163570)({
 Type: (string) (len=6) "string",
 Description: (string) "",
 Value: (int) 11,
 Required: (bool) false,
 Default: (interface {}) <nil>,
 Status: (string) "",
 Schema: (interface {}) <nil>,
 multiline: (bool) true
})
```  

Also, here the execution continues as there is no check on errorParser and return if errorParser is not `<nil>`

mrutkows: Can we do this on a separate issue?  I would like to get this merged for the tests included thus far.  If we try to address all the TODOs I added regarding error/warning handling this issue will get too large IMO. @pritidesai

pritidesai: sure, sounds good to me, thanks

Issue #296, opened 2017-07-19
Invalid Deployment. yaml for use case "webhook"


mrutkows:
The deployment YAML contains references to "WskTool" (invalid deleted repo.). It also includes an Daisy's username as default value where we should not be including developer usernames if possible in test files.

```
pplication:
  name: wskdeploy-samples
  namespace: guest

  packages:
    git_webhook:
      credential: 23bc46b1-71f6-4ed5-8c54-816aa4f8c502:123zO3xZCLrMN6v2BKK1dXYFpXlPkccOFqm12CdAsMgRU4VrNZ9lyGVCGuMDGIwP

      feeds:
        webhook:
          #we treat the inputs as parameters besides --feed.
          inputs:
            username: daisy-ycguo 
            repository: https://github.com/openwhisk/wsktool.git
            accessToken:
            events: PUSH
            endpoint: openwhisk.ng.bluemix.net
```

In addition, we should try not to use "guest" namespace (but use '_') default namespace where possible to allow testing/deployment on multiple target providers of openwhisk.

lzbj:
@daisy-ycguo ,could you please help take a look when you available? This use case seems added by you.

daisy-ycguo:
@lzbj OK

mrutkows:
addressed this issue in https://github.com/apache/incubator-openwhisk-wskdeploy/pull/300

PR #300, opened 2017-07-24
fix issues in webhook use case

daisy-ycguo:
Address issues #296, including
- use placeholder instead of real data in deployment.yml
- use default namespace _ other than guest namespace in deployment.yml


tests/usecases/webhook/deployment.yml
```diff
@@ -12,8 +12,8 @@ application:
           inputs:
             # TODO(dliu): Please fix "repsotiory value" and see if we can change/remove username
```
mrutkows: @daisy-ycguo Please remove the TODO comment and add a new comment telling user how to change the "username" and "accesstoken" and "repository".  For this, you can use fake example names.

tests/usecases/webhook/deployment.yml
```diff
@@ -12,8 +12,8 @@ application:
...
             # as actual developer names should not be in these files when possible.
-            username: daisy-ycguo
-            repository: https://github.com/openwhisk/wsktool.git
-            accessToken:
-            events: PUSH
-            endpoint: openwhisk.ng.bluemix.net
```
mrutkows: @daisy-ycguo Please do not assume bluemix as the endpoint, but please add another comment here that uses bluemix as an example and also describes what the endpoint would look like if installed locally (native).

daisy-ycguo:
I will close it and create a new PR.

PR #310, opened 2017-08-09
Support both manifest.yaml and manifest.yml

lzbj:
Mitigate same variable copy around.


cmd/root.go
```diff
@@ -144,3 +138,134 @@ func initConfig() {
...
+	if utils.Flags.ManifestPath == "" {
+		if _, err := os.Stat(path.Join(projectPath, "manifest.yaml")); err == nil {
+			utils.Flags.ManifestPath = path.Join(projectPath, deployers.ManifestFileNameYaml)
+		} else if _, err := os.Stat(path.Join(projectPath, "manifest.yml")); err == nil {
+			utils.Flags.ManifestPath = path.Join(projectPath, deployers.ManifestFileNameYml)
+		}
+	}
+
+	if utils.Flags.DeploymentPath == "" {
+		if _, err := os.Stat(path.Join(projectPath, "deployment.yaml")); err == nil {
```
mrutkows: @lzbj currently, we have 2 hardcoded filenames for manifest and for deployment files (i.e., manifest.yaml and .yml and deployment.yaml and .yml respectively (declared in "shared.go").  These names appear "lowercased". We need to find a way such that we can match any mixed base name such as "Manifest.yaml" or "manifest.YAML" etc.  I would be happy if you want to open this as a lower priority issue to fix later.

cmd/root.go
```diff
@@ -144,3 +138,134 @@ func initConfig() {
...
+			utils.Flags.DeploymentPath = path.Join(projectPath, deployers.DeploymentFileNameYaml)
+		} else if _, err := os.Stat(path.Join(projectPath, "deployment.yml")); err == nil {
+			utils.Flags.DeploymentPath = path.Join(projectPath, deployers.DeploymentFileNameYml)
+		}
+	}
+
+	if utils.MayExists(utils.Flags.ManifestPath) {
```
mrutkows: @lzbj we need to add testcases to test for the "-p" parameter for a package with 
1) only a manifest.yaml (and .yml)
2) a manifest.yaml and deployment.yml (both)
3) mix of .yaml / .yml combinations
and automate in travis.

mrutkows: @lzbj Need to also add a testcase for the <path> to the package that ends with a trailing "/" slash and another without.

cmd/root.go
```diff
@@ -144,3 +138,134 @@ func initConfig() {
...
+		deployer.IsDefault = utils.Flags.UseDefaults
+
+		userHome := utils.GetHomeDirectory()
+		propPath := path.Join(userHome, ".wskprops")
+
+		whiskClient, clientConfig := deployers.NewWhiskClient(propPath, utils.Flags.DeploymentPath, deployer.IsInteractive)
+		deployer.Client = whiskClient
+		deployer.ClientConfig = clientConfig
+
+		verifiedPlan, err := deployer.ConstructUnDeploymentPlan()
```
mrutkows: @lzbj before we "deploy" when a "-p" package path is provided, we should (in verbose mode) tell the user (echo back) the path/filename we found and parsed/using to deploy.  We also need to have a clear error (with special text) that tells used "Manifest file not found at path 'x'".

cmd/root.go
lzbj: Agree, I will open a issue for it.

cmd/root.go
lzbj: Ok, I'll try add some unitary test code.

cmd/root.go
lzbj: Ok, agree.

lzbj:
@mrutkows , added test cases based on your comments, please review.

mrutkows:
@lzbj @priti we have reason to believe that all the integration tests are only running up until they get ready to deploy (i.e., only testing the parser) and the query "deploy Y/N" somehow always answers "no" and that is why tests may be reporting PASS when they fail to deploy.

tests/src/integration/common/wskdeploy.go
```diff
@@ -18,38 +18,46 @@
...
-    return wskdeploy.RunCommand("undeploy", "-m", manifestPath, "-d", deploymentPath)
+	return wskdeploy.RunCommand("undeploy", "-m", manifestPath, "-d", deploymentPath)
+}
+
+func (wskdeploy *Wskdeploy) DeployProjectPathOnly(projectPath string) ([]byte, error) {
+        return wskdeploy.RunCommand("-p", projectPath)
+}
+
+func (wskdeploy *Wskdeploy) DeployManifestPathOnly(manifestpath string) ([]byte, error) {
+	return wskdeploy.RunCommand("-p", manifestpath)
```
pritidesai: this should be 

```
return wskdeploy.RunCommand("-m", manifestpath)
```

instead of 

```
return wskdeploy.RunCommand("-p", manifestpath)
```

Integration tests are not clearly catching this bug, when I run it manually:

```
./wskdeploy -p tests/src/integration/flagstests/manifest.yaml 
2017/08/17 15:46:51 Manifest file not found at path /Users/pritidesai/Documents/goworkspace/src/github.com/apache/incubator-openwhisk-wskdeploy/tests/src/integration/flagstests/manifest.yaml
```

But two integration tests we have which are using this method (1) TestSupportManifestYamlPath (2) TestSupportManifestYmlPath are passing (PASS) on Travis.

lzbj: @pritidesai , thanks for your review, I'll change the test code accordingly. 
@mrutkows , we could set the interactive mute, so the deploy will go by default. I'll change the test code.

mrutkows: @lzbj @pritidesai what issue do you want to track this issue on as we decided to merge this PR?

PR #314, opened 2017-08-10
Pull the openwhisk images instead of building them in Travis

houshengbo:


mrutkows:
@houshengbo @daisy-ycguo thank you very much!

PR #370, opened 2017-08-21
Assure openwhisk install bash exits immediately on failure.

mrutkows:


daisy-ycguo:
PR #371  is to fix the same issue and has been merged. I will close this one then.

Issue #360, opened 2017-08-17
Validate integration tests are running as part of Travis build


pritidesai:
See issue #313, creating web action is failing but Travis build never complains about any failure running [webaction](https://github.com/apache/incubator-openwhisk-wskdeploy/tree/master/tests/src/integration/webaction) integration test.

pritidesai:
I just ran a dump on one of the integration tests and found that the deployment is being canceled as there is no "Y" provided on cmdline while executing wskdeploy: 

```
go test -v ./tests/src/integration/zipaction/ -tags=integration
=== RUN   TestZipAction
([]uint8) (len=587 cap=1536) {
 00000000  20 20 20 20 20 20 20 20  20 5f 5f 5f 5f 20 20 20  |         ____   |
 00000010  20 20 20 5f 5f 5f 20 20  20 20 20 20 20 20 20 20  |   ___          |
 00000020  20 20 20 20 20 20 20 20  20 5f 20 20 20 20 5f 20  |         _    _ |
 00000030  5f 20 20 20 20 20 5f 20  20 20 20 20 5f 0a 20 20  |_     _     _.  |
 00000040  20 20 20 20 20 20 2f 5c  20 20 20 5c 20 20 20 20  |      /\   \    |
 00000050  2f 20 5f 20 5c 20 5f 20  5f 5f 20 20 20 5f 5f 5f  |/ _ \ _ __   ___|
 00000060  20 5f 20 5f 5f 20 7c 20  7c 20 20 7c 20 7c 20 7c  | _ __ | |  | | ||
 00000070  5f 5f 20 28 5f 29 5f 5f  5f 7c 20 7c 20 5f 5f 0a  |__ (_)___| | __.|
 00000080  20 20 20 2f 5c 20 20 2f  5f 5f 5c 20 20 20 5c 20  |   /\  /__\   \ |
 00000090  20 7c 20 7c 20 7c 20 7c  20 27 5f 20 5c 20 2f 20  | | | | | '_ \ / |
 000000a0  5f 20 5c 20 27 5f 20 5c  7c 20 7c 20 20 7c 20 7c  |_ \ '_ \| |  | ||
 000000b0  20 27 5f 20 5c 7c 20 2f  20 5f 5f 7c 20 7c 2f 20  | '_ \| / __| |/ |
 000000c0  2f 0a 20 20 2f 20 20 5c  5f 5f 5f 5f 20 5c 20 20  |/.  /  \____ \  |
[...]

PR #372, opened 2017-08-22
Integration Tests

pritidesai:
Making sure that we are running integration tests as part of Travis.

Updated all the existing integration tests to reflect changes in the underlying methods including its return type.


daisy-ycguo:
@houshengbo The PR is to improve integration test and address issue #360 . Will you review it ?

daisy-ycguo:
When issue #374 is fixed, we can disable interative mode, other than sending a Y character to the stdin.

pritidesai:
Thanks @daisy-ycguo I have comments in the code itself, to improve on sending "y" on STDIN.

pritidesai:
@daisy-ycguo @lzbj @houshengbo all the integration tests are failing with `{"error":"Missing cmd key"}` what does this error mean? is `utils.Flags.WithinOpenWhisk`  set to `true` in travis builds? wskdeploy still exits with 0 status which is odd.

lzbj:
@pritidesai @daisy-ycguo @houshengbo , use command return code to determine deploy failed or pass is one way  to help improve , but sometimes, even deploy failed, the command could not reflect what really happened or even return success.
To better address this, I suggest we could use wsk cli lib or rest calls to make sure the deploy behavior is what we expected.

pritidesai:
Have decided to fix every failed integration tests in separate PR, still need some code from this PR, specially from wskdeploy.go, I had added timeout enhancement here to  quit "wskdeploy" execution if it takes more than 30 seconds(configurable).

Issue #374, opened 2017-08-22
Interactive mode should be configurable


houshengbo:
We need to make the interactive mode of wskdeploy configurable, so that we do not have to enter "yes" all the time to issue a command.

mrutkows:
fixed in https://github.com/apache/incubator-openwhisk-wskdeploy/pull/375

PR #375, opened 2017-08-22
Make interative mode able to set to false

houshengbo:
The default value of interative mode should be false, instead of depending
on utils.Flags.WithinOpenWhisk.

Closes-Bug: #374

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -113,7 +113,7 @@ func init() {
 	RootCmd.Flags().StringVarP(&utils.Flags.ProjectPath, "pathpath", "p", ".", "path to serverless project")
 	RootCmd.Flags().StringVarP(&utils.Flags.ManifestPath, "manifest", "m", "", "path to manifest file")
 	RootCmd.Flags().StringVarP(&utils.Flags.DeploymentPath, "deployment", "d", "", "path to deployment file")
-	RootCmd.PersistentFlags().BoolVarP(&utils.Flags.UseInteractive, "allow-interactive", "i", !utils.Flags.WithinOpenWhisk, "allow interactive prompts")
+	RootCmd.PersistentFlags().BoolVarP(&utils.Flags.UseInteractive, "allow-interactive", "i", false, "allow interactive prompts")
 	RootCmd.PersistentFlags().BoolVarP(&utils.Flags.UseDefaults, "allow-defaults", "a", false, "allow defaults")
 	RootCmd.PersistentFlags().BoolVarP(&utils.Flags.Verbose, "verbose", "v", false, "verbose output")
 	RootCmd.PersistentFlags().StringVarP(&utils.Flags.ApiHost, "apihost", "", "", wski18n.T("whisk API HOST"))
--- a/deployers/servicedeployer.go
+++ b/deployers/servicedeployer.go
@@ -119,7 +119,7 @@ func (deployer *ServiceDeployer) ConstructDeploymentPlan() error {
 
 	manifestReader.InitRootPackage(manifestParser, manifest)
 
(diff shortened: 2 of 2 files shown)


cmd/root.go
```diff
@@ -113,7 +113,7 @@ func init() {
 	RootCmd.Flags().StringVarP(&utils.Flags.ProjectPath, "pathpath", "p", ".", "path to serverless project")
 	RootCmd.Flags().StringVarP(&utils.Flags.ManifestPath, "manifest", "m", "", "path to manifest file")
 	RootCmd.Flags().StringVarP(&utils.Flags.DeploymentPath, "deployment", "d", "", "path to deployment file")
-	RootCmd.PersistentFlags().BoolVarP(&utils.Flags.UseInteractive, "allow-interactive", "i", !utils.Flags.WithinOpenWhisk, "allow interactive prompts")
```
daisy-ycguo: I agree with this change.

deployers/servicedeployer.go
```diff
@@ -119,7 +119,7 @@ func (deployer *ServiceDeployer) ConstructDeploymentPlan() error {
 
 	manifestReader.InitRootPackage(manifestParser, manifest)
 
-	if deployer.IsDefault == true && !utils.Flags.WithinOpenWhisk {
+	if deployer.IsDefault == true {
```
daisy-ycguo: I hope to include David to review this line of change.

deployers/servicedeployer.go
```diff
@@ -183,7 +183,7 @@ func (deployer *ServiceDeployer) ConstructUnDeploymentPlan() (*DeploymentApplica
 // TODO(TBD): according to some planning?
 func (deployer *ServiceDeployer) Deploy() error {
 
-	if deployer.IsInteractive == true && !utils.Flags.WithinOpenWhisk {
+	if deployer.IsInteractive == true {
```
daisy-ycguo: I hope to include David to review this line of change.

pritidesai:
LGTM,  thanks!

deployers/servicedeployer.go
lzbj: thanks @daisy-ycguo , LGTM.

deployers/servicedeployer.go
```diff
@@ -119,7 +119,7 @@ func (deployer *ServiceDeployer) ConstructDeploymentPlan() error {
 
 	manifestReader.InitRootPackage(manifestParser, manifest)
 
-	if deployer.IsDefault == true && !utils.Flags.WithinOpenWhisk {
```
mrutkows: I am curious as why we were testing "WithinOpenWhisk" here in the first place...

pritidesai: @lionelvillard had added this flag to indicate that the wskdeploy is being run as an OpenWhisk action itself. If it is being run as an OpenWhisk action than skip this whole block of execution where wskdeploy is reading and inspecting project directory where action files are located, reading content of the action files and setting exec.Code of an ActionRecord. Hope this helps. We cannot set exec.Code this way if wskdeploy is running as an OpenWhisk action as none of the directory structure is accessible in cloud. But than I am trying to figure out myself what is the use of having wskdeploy run as an OpenWhisk action.

davidbreitgand: With the new "export" capability added by @kpavel, the directory structure will be exported to the FS of the container, in which the action runs. The use case would be: "export" assets from one OW instance and "import" into another one via running wskdeploy from within an action.

davidbreitgand: With the new "export" capability added by @kpavel, the directory structure will be exported to the FS of the container, in which the action runs. The use case would be: "export" assets from one OW instance and "import" into another one via running wskdeploy from within an action.

Issue #305, opened 2017-08-07
Wondering why triggers are updated with overwrite set to false


zachschultz:
I have a wskdeploy manifest yaml file that looks like:

```
package:
  name: messagehub-trigger
  version: 1.0
  license: Apache-2.0
  namespace: _
  dependencies:
    openwhisk-messagehub:
      location: /whisk.system/messaging
      inputs:
        kafka_brokers_sasl: $KAFKA_BROKERS
        user: $MESSAGEHUB_USER
        password: $MESSAGEHUB_PASS
        kafka_admin_url: $KAFKA_ADMIN_URL
  actions:
    process-message:
      location: actions/process-message.js
  triggers:
    message-received-trigger:
      source: openwhisk-messagehub/messageHubFeed
      inputs:
        isJSONData: true
        topic: $KAFKA_TOPIC
  rules:
      log-message-rule:
        trigger: message-received-trigger
        action: messagehub-trigger/process-message
```

When I changed the `source` of the `message-received-trigger` and ran wskdeploy I noticed that OpenWhisk didn't actually change the trigger, and after running with `verbose` I saw that the request was:

```
REQUEST:
[PUT]	https://openwhisk.ng.bluemix.net:443/api/v1/namespaces/_/triggers/message-received-trigger?overwrite=false
```

Why is it that everything else is sent with `overwrite=true` but trigger creations are sent with `overwrite=false`?

zachschultz:
It looks like the relevant code is here and that trigger feed actions are treated differently: https://github.com/apache/incubator-openwhisk-wskdeploy/blob/a67e7285845828510c2c6c8849c950c37ca4576d/deployers/servicedeployer.go#L415

csantanapr:
I think the wskdeploy needs to do more with feed actions to be able to updated them, it needs to delete the trigger and invoke the feed action with the with delete lifecycle, and then follow with normal feed creation.

daisy-ycguo:
@zachschultz You are right, I think it should be set to false.
```
_, _, err := deployer.Client.Triggers.Insert(t, false)
```

daisy-ycguo:
@zachschultz There is a PR #377 to address this issue. But I'm very curious to your case. What's the new value you want to update the `source` of the `message-received-trigger`?

zachschultz:
[...]

PR #377, opened 2017-08-23
Allow trigger update when creating triggers

daisy-ycguo:
When creating a feed trigger, set 'overwrite=true'
Address issue #305

### DIFF
--- a/deployers/servicedeployer.go
+++ b/deployers/servicedeployer.go
@@ -412,7 +412,7 @@ func (deployer *ServiceDeployer) createFeedAction(trigger *whisk.Trigger, feedNa
 		Publish:     &pub,
 	}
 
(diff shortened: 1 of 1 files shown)


deployers/servicedeployer.go
```diff
@@ -412,7 +412,7 @@ func (deployer *ServiceDeployer) createFeedAction(trigger *whisk.Trigger, feedNa
 		Publish:     &pub,
 	}
 
-	_, _, err := deployer.Client.Triggers.Insert(t, false)
+	_, _, err := deployer.Client.Triggers.Insert(t, true)
```
mrutkows: Defaulting to override=true as the default matches all the other Insert() calls for all other entities.

PR #379, opened 2017-08-23
Remove the explicit api keys to access kafka service

houshengbo:


houshengbo:
@daisy-ycguo @lzbj 
Please be advised that do not put valid keys in the repos.

PR #397, opened 2017-08-28
Restore the integration tests

houshengbo:
In the new way to load the credentials, we are now able to run the
following integration tests:
* jaractionTest
* triggerrulTest
* zipactionTest
* flagsTest

We have to adopt a new model to create the integration tests and run
them in a consolidated test file, in order to avoid the concurrent
issues of accessing the same resources in openwhisk at the same time.

Closes-Bug: #398


houshengbo:
@markusthoemmes @pritidesai 
You will love this PR.

Issue #417, opened 2017-08-30
New output messages from wskdeply


daisy-ycguo:
Do we really want to print these messages ? Or they are debug messages which should be removed ?

```
$ ./wskdeploy -i -m ~/workspace/owbp-cloudant-trigger/runtimes/node/manifest.yaml
2017/08/30 06:12:53 Unmarshal Openwhisk info from internet
2017/08/30 06:12:53 The URL is https://openwhisk.ng.bluemix.net/api, selected from .wskprops
2017/08/30 06:12:53 The auth key is set, selected from .wskprops
2017/08/30 06:12:53 The namespace is _, selected from manifest.yml
```

mrutkows:
As part of Vincent's work I hope we have something like a "printWarning()" printError() printInfo() printTrace() in a shared package/module (with standard formatting that allow timestamp, module, line number for trace and YAML file/line number if parser finds an error in a YAML file.  We can do what the CLI does and print trace if an Environment Variable is set to value > 0.  @houshengbo

mrutkows:
If we have ^^^ then we can decide which ones are info, verbose and which are trace...  and turn on verbose when user sets flag, turn on trace if env. variable is set to non-zero (true) value.

mrutkows:
@daisy-ycguo @pritidesai @houshengbo for now the more messages the better as we work down our backlog and we can move most to a printTrace() or printVerbose() type method later.

daisy-ycguo:
Close it as it's not valid now.

Issue #340, opened 2017-08-16
Create Error handling /display path through all code


mrutkows:
Many errors/warnings are captured but not surfaced properly or reported; some are caught by a general "panic" method, but does not provide context of error.  Warnings are not surfaced at all.

These all need to be surfaced in a way such that "unwinding of a Deployment can happen properly.

Also need to consider i18n messages and specific named Error types with detailed messages. 

If parsing errors are found the line number needs to be included,
If deployment (API) errors are encountered we should display detailed request/response/payload information.

mrutkows:
See https://github.com/apache/incubator-openwhisk-wskdeploy/issues/224 for spec. impacts.

PR #420, opened 2017-08-30
Add the error handling mechanism

houshengbo:
This PR makes the following changes:
* Replace log with fmt, since wskdeploy is a client tool, which does not need t$
information.
* In test cases, read the stardard output and the standard error separately. We$
parse the standard error to determine if the execution of the a command succeed$

Closes-Bug: #340 #417


daisy-ycguo:
Related with #417

mrutkows:
@daisy thanks for linking as it has notes that affect how we would code this.

Issue #415, opened 2017-08-30
dependent packages requires manifest.yml and doesn't support manifest.yaml


pritidesai:
wskdeploy does not support dependent packages which has manifest.yaml file. It only deploys dependent package with manifest.yml. Please add support for manifest.yaml. Refer to integration test on dependency for more info: https://github.com/apache/incubator-openwhisk-wskdeploy/tree/master/tests/src/integration/dependency

lzbj:
@pritidesai , I can't reproduce this issue, could you help confirm.
I forked your repo here.https://github.com/lzbj/openwhisk-app
then I changed the yaml like this:
diff --git a/tests/src/integration/dependency/manifest.yaml b/tests/src/integration/dependency/manifest.yaml
index f76390d..e85bc4b 100644
--- a/tests/src/integration/dependency/manifest.yaml
+++ b/tests/src/integration/dependency/manifest.yaml
@@ -2,7 +2,7 @@ package:
   name: TestGitHubDependency
   dependencies:
     openwhisk-app:
-      location: github.com/pritidesai/openwhisk-app
+      location: github.com/lzbj/openwhisk-app
   triggers:
     trigger1:


THen I run the deploy.

2017/08/30 14:15:18 Unmarshal Openwhisk info from internet
2017/08/30 14:15:18 The URL is https://openwhisk.ng.bluemix.net/api, selected from .wskprops
2017/08/30 14:15:18 The auth key is set, selected from .wskprops
2017/08/30 14:15:18 The namespace is david.liu@cn.ibm.com_dliu, selected from .wskprops
2017/08/30 14:15:21 Deploying package TestGitHubDependency ... 
2017/08/30 14:15:22 Done!
Deploying dependency openwhisk-app ... 
2017/08/30 14:15:22 Deploying package openwhisk-app ... 
2017/08/30 14:15:23 Done!
2017/08/30 14:15:23 Deploying action openwhisk-app/helloworld ... 
2017/08/30 14:15:24 Done!
2017/08/30 14:15:24 Deploying action openwhisk-app/hello ... 
2017/08/30 14:15:24 Done!
Done!
Done!
Done!
Deploying rule rule1 ... Done!
Deploying rule rule2 ... Done!
2017/08/30 14:15:26 
Deployment completed successfully.

lzbj:
I have already changed the yaml file in my repo to .yaml, did I miss some steps to reproduce?

pritidesai:
[...]

PR #424, opened 2017-08-31
Use const value to replace hard coded manifest yaml file name.

lzbj:


lzbj:
For fix issue https://github.com/apache/incubator-openwhisk-wskdeploy/issues/415

Issue #356, opened 2017-08-16
Integration Test - hello world


pritidesai:
Create an integration test based on manifest in [helloworld under use cases](https://github.com/apache/incubator-openwhisk-wskdeploy/tree/master/tests/usecases/helloworld). Add hello world actions in all supported runtimes in addition to nodejs.

Create integration tests under test/src/integration.

If this manifest file is referenced in any of the test cases, update the test to read the manifest file from [tests/dat](https://github.com/apache/incubator-openwhisk-wskdeploy/tree/master/tests/dat)

pritidesai:
I am working on this and soon create a PR. Thanks!

pritidesai:
Closing this as its fixed with PR #432

PR #431, opened 2017-09-01
Adding integration test on helloworld

pritidesai:
Closes issue #356


pritidesai:
creating updated PR with latest master

PR #432, opened 2017-09-01
Adding integration test on helloworld

pritidesai:
Closes issue #356


houshengbo:
@pritidesai Where are we going to put the manifest and deployment? Is it under usecase or integration folder?

pritidesai:
@houshengbo its under integration directory, as we are creating integration test for hello world app.

```
tests/src/integration/helloworld/deployment.yaml
tests/src/integration/helloworld/manifest.yaml
```

The one under use cases folder is deleted as its being converted to integration test. The apps or packages which can't have integration test, goes under use cases folder.

Issue #422, opened 2017-08-30
Validate all entity names <= 256 chars before create/update to server


mrutkows:
For all "entity names" (i.e., Actions, Triggers, Packages, Rules, etc.) we should validate (on the client side) that they are less than or equal to 256 characters (including their full namespaced named).

• For example, the Action name (i.e., <actionName> MUST be less than or equal to 256 characters.

Note: THis include the TOTAL length of the namespace; therefore we MUST make the length calculation include the FULL name including namespace

PR #435, opened 2017-09-01
deprecating swift and php runtimes

### LINKED ISSUE #429: runtime swift vs swift:2 vs swift:3.1.1
I am writing an action and have set runtime to `swift` or `swift:2` which fails:

```
2017/08/31 14:09:46 Got error creating action with error message: The 'swift' runtime is no longer supported. You may read and delete but not update or invoke this action. (code 196) and error code: 144.
```
This error disappears and action is deployed when runtime is set to `swift:3.1.1`. 

Does openwhisk support swift:2? One of the blog posts says "If you have any Swift v2 actions, you will want to migrate those to the Swift v3 runtime."

pritidesai:
we have swift:3, swift:3.1.1, and php:7.1 supported.

Implicit runtimes `swift` and `php` are not supported, replacing them with `swift:3` and `php:7.1`.

Action creation fails with this message when implicit runtime is set to `php`:

```
"error":"The request content was malformed:\nkind 'php' not in Set(php:7.1, swift:3, nodejs, blackbox, java, sequence, nodejs:6, python:3, python, python:2, swift, swift:3.1.1)","code":9942234}
```

Closes #429

### DIFF
--- a/parsers/manifest_parser.go
+++ b/parsers/manifest_parser.go
@@ -267,7 +267,7 @@ func (dm *YAMLParser) ComposeActions(mani *ManifestYAML, manipath string) (ar []
 
 				switch ext {
 				case ".swift":
-					kind = "swift"
+					kind = "swift:3"
 				case ".js":
 					kind = "nodejs:6"
 				case ".py":
@@ -275,7 +275,7 @@ func (dm *YAMLParser) ComposeActions(mani *ManifestYAML, manipath string) (ar []
 				case ".java":
 					kind = "java"
 				case ".php":
-					kind = "php"
+					kind = "php:7.1"
 				case ".jar":
 					kind = "java"
 				default:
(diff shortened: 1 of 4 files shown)


mrutkows:
@pritidesai  the goal was to all simply "swift" to be mapped to the "latest" swift image and "php" to be mapped to the latest php image.  This would require us to provide a mapping based upon the (no hardcoded) JSON the platform returns and performing a lookup into it to determine "latest".

mrutkows:
@pritidesai I am happy if this gets things working for now,  but we should open a lower priority issue to add the "being nice" features I described above.

pritidesai:
@mrutkows yup this is very quick and dirty fix to have wskdeploy work with `php` and `swift` runtimes. 

I have created issue #422 to add warnings and #443 to add such flag.

Issue #327, opened 2017-08-15
Unit Test - yamlparser.go


pritidesai:
Add unit tests to validate methods under parsers package which are defined in [yamlparser.go](https://github.com/apache/incubator-openwhisk-wskdeploy/blob/master/parsers/yamlparser.go).

All unit tests should go in `yamlparser_test.go` under [parsers/](https://github.com/apache/incubator-openwhisk-wskdeploy/blob/master/parsers)

```
// Compose methods
// compose trigger
func (trigger *Trigger) ComposeWskTrigger(kvarr []whisk.KeyValue) *whisk.Trigger {
...
}

// compose rule
func (rule *Rule) ComposeWskRule() *whisk.Rule {
...
}

// compose package
func (pkg *Package) ComposeWskPackage() *whisk.Package {
...
}

// get list of actions
func (pkg *Package) GetActionList() []Action {
...
}

// get list of triggers
func (pkg *Package) GetTriggerList() []Trigger {
...
}

// get list of rules
func (pkg *Package) GetRuleList() []Rule {
...
}

// get list of feeds
func (pkg *Package) GetFeedList() []Feed {
...
}
```

mrutkows:
Fixed in 440

PR #440, opened 2017-09-05
more unit tests for yamlparser.go

daisy-ycguo:
Addressed issue #327


pritidesai:
thanks @daisy-ycguo LGTM

Issue #322, opened 2017-08-15
Unit Test - manifest_parser.go


pritidesai:
Add unit tests to validate methods under `parsers` package which are defined in [manifest_parser.go](https://github.com/apache/incubator-openwhisk-wskdeploy/blob/master/parsers/manifest_parser.go).

All unit tests should go in `manifest_parser_test.go` under [parsers/](https://github.com/apache/incubator-openwhisk-wskdeploy/blob/master/parsers)

```
// Read existing manifest file or create new if none exists
func ReadOrCreateManifest() *ManifestYAML {
...
}

// Serialize manifest to local file
func Write(manifest *ManifestYAML, filename string) {
...
}

// deserialize manifest file into ManifestYAML struct
func (dm *YAMLParser) Unmarshal(input []byte, manifest *ManifestYAML) error {
...
}

// parse manifest file into ManifestYAML struct
func (dm *YAMLParser) ParseManifest(mani string) *ManifestYAML {
...
}

// compose dependencies and set DependencyRecord
func (dm *YAMLParser) ComposeDependencies(mani *ManifestYAML, projectPath string) (map[string]utils.DependencyRecord, error) {
...
}

// compose package and set whisk.Package 
func (dm *YAMLParser) ComposePackage(mani *ManifestYAML) (*whisk.Package, error) {
...
}

// compose sequences and set ActionRecord 
func (dm *YAMLParser) ComposeSequences(namespace string, mani *ManifestYAML) ([]utils.ActionRecord, error) {
...
}

// compose Action and set ActionRecord
func (dm *YAMLParser) ComposeActions(mani *ManifestYAML, manipath string) (ar []utils.ActionRecord, aub []*utils.ActionExposedURLBinding, err error) {
...
}

// compose Triggers and set whisk.Trigger
func (dm *YAMLParser) ComposeTriggers(manifest *ManifestYAML) ([]*whisk.Trigger, error) {
...
}

// compose Rules and set whisk.Rule 
func (dm *YAMLParser) ComposeRules(manifest *ManifestYAML) ([]*whisk.Rule, error) {
...
}

// Resolve input parameter with type, value, or default value
func ResolveParameter(paramName string, param *Parameter) (interface{}, error) {
...
}


```

PR #441, opened 2017-09-05
more unit tests for manifest_parser.go

daisy-ycguo:
Addressed issue #322.


pritidesai:
thanks @daisy-ycguo LGTM

Issue #437, opened 2017-09-05
Create binding when dependency name and the origin package name are different


daisy-ycguo:
Below are `dependencies` samples. 
```
dependencies: 
    myCloudant:
      location: /whisk.system/cloudant
      inputs:
        dbname: myGreatDB
   myhelloworld:
      location: github.com/daisy-ycguo/wskdeploy-test/helloworld
```

Now we can define two kinds of dependencies:
- Binding to existing package, e.g. `myCloudant` in the sample.  We will create a package binding to `/whisk.system/cloudant` with the new package name as `myCloudant`.
- Dependency to an external source code in git repo, e.g. `myhelloworld` in the sample. We will deploy the package `myhelloworld` using the manifest file under folder `http://github.com/daisy-ycguo/wskdeploy-test/helloworld`.

If the dependency name is different than the origin package name, the reference to actions under this package will look weird. For example, the origin package name is `myhelloworld`, the dependency name is `dependency_name`. When I want to refer to an action under this package, I have to use `myhelloworld/helloworld`, not `dependency_name/helloworld`. Refer to below sample.
```
dependencies: 
   dependency_name:
      location: github.com/daisy-ycguo/wskdeploy-test/helloworld
sequences:
    mySequence:
      actions: myhelloworld/helloworld
```

So my question is how to handle if the dependency name is different with the origin package name. Maybe we can deploy the origin package at first and then create a binding with the dependency name. It might be a reasonable solution.

@mrutkows what's your suggestion?

mrutkows:
@daisy-ycguo  The intent was that the  dependency's <package_name> would be treated as a local alias for the actual package name as decribed in the referenced package.  The referenced package would have its own Manifest file that would include its actual Package name (and the one that would be used by the wskdeploy tool to replace the local alias).

mrutkows:
@daisy-ycguo this would align with our decision on how to handle 'rules' and 'sequence' name references (that is, assume their names to be package-scoped names unless otherwise specified) in https://github.com/apache/incubator-openwhisk-wskdeploy/issues/394

daisy-ycguo:
I verified and I think package binding can help this use case.
I made following test.
1. create package `IntegrationTestHelloWorld` by `wskdeploy -p tests/src/integration/helloworld`
2. create a binding package newpkg by `wsk package bind IntegrationTestHelloWorld newpkg`, and then get the newpkg by `wsk package get --summary newpkg`, I see all the actions are listed:
[...]

PR #461, opened 2017-09-07
Create binding when dependency name is different

daisy-ycguo:
Addressed issue #437 

This PR includes
- create a binding when dependency name is different from the origin package name
- change usecase/dependency to show this capability
- a README.md for usecase/dependency


pritidesai:
LGTM (except conflict)

Thanks a bunch @daisy-ycguo for adding such enhancement, its working great, very happy with the results, thanks again!

mrutkows:
@daisy we have moved Paul's package to incubator-openwhisk-test repo. and we will want to update this use case to reference it there. In addition, we need to move your package there too so there are no private repo. names that are not part of the Apache GitHub referenced.

Issue #450, opened 2017-09-06
Trigger supports "source" according to the code, yet spec has "feed" as trigger sub element.


daisy-ycguo:
We should support `feed` under `trigger`, mark `source` deprecated.

In the use case `alarmtrigger`, a trigger is described as below.
```
package:
    name: helloworld
    triggers:
        Every12Hours:
            source: /whisk.system/alarms/alarm
```
If I change `source` to `feed`, it will create a trigger without `feed` in OpenWhisk.
```
$ wsk trigger get Every12Hours
ok: got trigger Every12Hours
{
    "namespace": "guoyingc@cn.ibm.com_dev",
    "name": "Every12Hours",
    "version": "0.0.1",
    "parameters": [
        {
            "key": "cron",
            "value": "0 */12 * * *"
        }
    ],
    "limits": {},
    "publish": false
}
```

mrutkows:
@daisy-ycguo agree. we should support the spec. keyname "feed" and deprecate "source" as we did recently for the "location" key under the Action entity.

daisy-ycguo:
Fixed and closed.

PR #465, opened 2017-09-08
Use feed in trigger entity instead of source

daisy-ycguo:
Addressed issue #450 

In this PR, below updates are made:
- Change the code to support 'feed' in trigger entity; mark 'source' as deprecated.
- Print warning messages if 'source' is being used. 
```
WARNING: The 'source' YAML key in trigger entity will soon be deprecated. Please use 'feed' instead as described in specifications.
```
- Change the 'source' to 'feed' in the manifest files in test cases

### DIFF
--- a/parsers/manifest_parser.go
+++ b/parsers/manifest_parser.go
@@ -31,6 +31,7 @@ import (
 	"fmt"
 	"github.com/apache/incubator-openwhisk-client-go/whisk"
 	"github.com/apache/incubator-openwhisk-wskdeploy/utils"
+	"github.com/apache/incubator-openwhisk-wskdeploy/wski18n"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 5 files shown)


parsers/manifest_parser.go
```diff
@@ -384,12 +384,20 @@ func (dm *YAMLParser) ComposeTriggers(manifest *ManifestYAML) ([]*whisk.Trigger,
 		pub := false
 		wsktrigger.Publish = &pub
 
-		keyValArr := make(whisk.KeyValueArr, 0)
+		//print warning information when .Source is not empty
 		if trigger.Source != "" {
+			fmt.Println("WARNING: The 'source' YAML key in trigger entity is deprecated. Please use 'feed' instead as described in specifications.")
```
mrutkows: @daisy-ycguo can we please use the new Debug method from misc.go? 

e.g.,:
        ```whisk.Debug(whisk.DbgWarn, warningString)``` ?

and please move the warning string to our translation file: ```wski18n/resources/en_US.all.json```

PR #476, opened 2017-09-12
WIP : Fixing the precedence order of retreiving credentials and adding unit tests

### LINKED ISSUE #380: WIP: Add unit tests to verify that wskdeploy is able to load credentials
We have a plan to follow precedence order of setting whisk credentials:

- API host for the OpenWhisk deployment

- API key for the subject invoking the action

- Namespace for the activation

Here is the precedence order:

(1) Deployment/Manifest file
(2) whisk properties / wskprops file
(3) Environment variables

Add verification in wskdeploy to verify that these credentials are being set before it attempts to run deployment/undeployment.

### LINKED ISSUE #471: OpenWhisk Credentials on Client Side should not parse whisk.properties
(Repeating Matt's comment from PR #468:) 

The order for credentials should be:

(1) Command line... (`wskdeploy --apihost <host> --auth <auth> --namespace <namespace>`) should override anything in static Deployment or Manifest YAML files. That is, the user has taken a manual action to supply the latest values.
(2) Deployment file overrides Manifest values
(3) Last default is Manifest file
(4) .wskprops, (NOT whisk.properties, this SHOULD NEVER be used directly). See https://github.com/apache/incubator-openwhisk/blob/master/docs/cli.md, Here it is ONLY used by someone setting up .wskprops to help them locate the APIHOST. It should NEVER be used by the CLI or wskdeploy directly.
(5) If interactive mode is enabled, then we can prompt for any missing (required) values.

**Changes needed:** 

Currently Credentials are loaded in the following precedence order:
(1) deployment file
(2) manifest file
(3) wskdeploy command line
(4) .wskprops
(5) whisk.properties

Change this order and ether to order from Matt's comment.

Update the unit tests (PR #468) based on this ordering.

pritidesai:
closes #380 
closes #471


deployers/whiskclient.go
```diff
@@ -20,194 +20,220 @@ package deployers
...
+		host, err := promptForValue("\nPlease provide the hostname for OpenWhisk [default value is openwhisk.ng.bluemix.net]: ")
+		utils.Check(err)
+		if host == "" {
+			host = "openwhisk.ng.bluemix.net"
+		}
+		apiHost.Value = host
+		apiHost.Source = INTERINPUT
+	}
+
+	if len(credential.Value) == 0 && isInteractive == true {
```
mrutkows: @pritidesai it seems we have many places where we use the "len" function to test for empty string; however, it seems GoLang convention is to simple test is  ```if mystring == "" ```.  Is there some downside to testing against a string literal (which should be faster)?

deployers/whiskclient.go
```diff
@@ -20,194 +20,220 @@ package deployers
...
+			mm := parsers.NewYAMLParser()
+			deployment, _ := mm.ParseDeployment(deploymentPath)
+			credential = GetPropertyValue(credential, deployment.Application.Credential, path.Base(deploymentPath))
+			namespace = GetPropertyValue(namespace, deployment.Application.Namespace, path.Base(deploymentPath))
+			apiHost = GetPropertyValue(apiHost, deployment.Application.ApiHost, path.Base(deploymentPath))
+		}
+	}
+
+	// read credentials from manifest file as didn't find them on command line and in deployment file
+	if len(credential.Value) == 0 || len(namespace.Value) == 0 || len(apiHost.Value) == 0 {
```
mrutkows: len() question below...

deployers/whiskclient.go
```diff
@@ -20,194 +20,220 @@ package deployers
...
+
+	clientConfig = &whisk.Config{
+		AuthToken: credential.Value, //Authtoken
+		Namespace: namespace.Value,  //Namespace
+		Host:      apiHost.Value,
+		Version:   "v1",
+		Insecure:  true, // true if you want to ignore certificate signing
+	}
+
+	if len(credential.Value) == 0 || len(apiHost.Value) == 0 || len(namespace.Value) == 0 {
```
mrutkows: len() question again

deployers/whiskclient.go
```diff
@@ -20,194 +20,220 @@ package deployers
...
-var promptForValue = func (msg string) (string) {
+var promptForValue = func(msg string) (string, error) {
 	reader := bufio.NewReader(os.Stdin)
 	fmt.Print(msg)
 
 	text, _ := reader.ReadString('\n')
 	text = strings.TrimSpace(text)
 
-	return text
+	return text, nil
```
mrutkows: Do we anticipate errors from these library calls?  If so, how do we actually catch them here and return a Whisk error?

tests/dat/deployment_validate_credentials.yaml
```diff
@@ -0,0 +1,11 @@
...
+# this is used for testing whiskclient functionality
+application:
+  name: UnitTestCredentials
+  apiHost: sample.deployment.openwhisk.org
+  credential: sample-deployment-credential
+  namespace: sample-deployment-namespace
+  packages:
+    ValidateWhiskConfig:
+      actions:
+        helloworld:
```
mrutkows: add another empty line at the end please

tests/dat/manifest_validate_credentials.yaml
```diff
@@ -0,0 +1,21 @@
...
+         name:
+           type: string
+           description: name of a person
+         place:
+           type: string
+           description: location of a person
+       outputs:
+         payload:
+           type: string
+           description: a simple greeting message, Hello World!
```
mrutkows: Add another line at the end please.

deployers/whiskclient.go
pritidesai: Both `len(string) == 0` and `string == ""` are idiomatic ways of checking if string is empty. To find which option is faster and results showed `len()` is faster (for fun :) as the length of a string is stored in the variable itself during declaration.

```
package main

import (
    "fmt"
    "time"
    "github.com/davecgh/go-spew/spew"
)

func main() {

    // Length is 3.
    string1 := "cat"
    // Length is 13 times 5 which is 65.
    string2 := "cat0123456789" +
        "cat0123456789" +
        "cat0123456789" +
        "cat0123456789" +
        "cat0123456789"

    spew.Dump(string1)
    spew.Dump(string2)

    t0 := time.Now()

    // Version 1: length of short string.
    for i := 0; i < 10000000; i++ {
        result := len(string1)
        if result != 3 {
            return
        }
    }

    t1 := time.Now()

    // Version 2: length of long string.
    for i := 0; i < 10000000; i++ {
        result := len(string2)
        if result != 65 {
            return
        }
    }

    t2 := time.Now()
    // Print results.
    fmt.Println(t1.Sub(t0))
    fmt.Println(t2.Sub(t1))
}
``` 

Results:

$ go run  main.go
(string) (len=3) "cat"
(string) (len=65) "cat0123456789cat0123456789cat0123456789cat0123456789cat0123456789"
2.752905ms
2.761132ms

```
package main

import (
    "fmt"
    "time"
)

func main() {

    // Length is 3.
    string1 := "cat"
    // Length is 13 times 5 which is 65.
    string2 := "cat0123456789" +
        "cat0123456789" +
        "cat0123456789" +
        "cat0123456789" +
        "cat0123456789"

    t0 := time.Now()

    // Version 1: length of short string.
    for i := 0; i < 10000000; i++ {
        if string1 == "" {
            return
        }
    }

    t1 := time.Now()

    // Version 2: length of long string.
    for i := 0; i < 10000000; i++ {
        if string2 == "" {
            return
        }
    }

    t2 := time.Now()
    // Print results.
    fmt.Println(t1.Sub(t0))
    fmt.Println(t2.Sub(t1))
}
```

Results:

go run  main.go
2.91229ms
3.12793ms

deployers/whiskclient.go
```diff
@@ -20,194 +20,220 @@ package deployers
...
+		if host == "" {
+			host = "openwhisk.ng.bluemix.net"
+		}
+		apiHost.Value = host
+		apiHost.Source = INTERINPUT
+	}
+
+	if len(credential.Value) == 0 && isInteractive == true {
+		cred, err := promptForValue("\nPlease provide an authentication token: ")
+		utils.Check(err)
```
houshengbo: Remove utils.check(), because we pop up all the error to the command line, and get rid of this method.

deployers/whiskclient.go
houshengbo: IMO, we do not error no matter what is the input. If there is no input, "" will be the value.

houshengbo:
@pritidesai I did not see you change any of the existing tests to load credentials from .wskprops or other sources. If whisk.property is removed, we need to provide a default way for integration tests to load the credentials.

deployers/whiskclient.go
mrutkows: @pritidesai Nice!  IMO, performance is the end argument. Was just checking as I have seen on StackOverflow that the "voted" / preferred way is using the string literal compare on more than one thread.  Did not want some GoLang(ista?) person coming and criticizing our code...

pritidesai:
@houshengbo I have all the tests in https://github.com/apache/incubator-openwhisk-wskdeploy/pull/476/files#diff-ca71183f6ce74a411d0d8a225461502b

also, we did not remove whisk.property, we changed it such that `wskdeploy` produces warning when `whisk.properties` is used to load credentials, warning saying its primarily used by Travis for integration tests and should not be used for production deployments.

pritidesai:
please review PR #485 --- sorry for creating one more PR for this, I am very bad at resolving conflicts :(

Issue #206, opened 2017-03-26
Add a timeout to access the baseurl


houshengbo:
If the baseurl we are about to access is not valid, it will take a very long time to get the error response. Suggest to add a timeout for whiskclient to access the baseurl.

lzbj:
@houshengbo , are you still look at this issue, could you give a little more info in the issue description, so I can fix it if you not available?

houshengbo:
#481 fixed this bug.

PR #481, opened 2017-09-13
Add default http timeout

houshengbo:
This PR adds 30 seconds as the default http timeout.
Closes-Bug: #206


houshengbo:
We need to merge https://github.com/apache/incubator-openwhisk-client-go/pull/35/files first for this patch.

Issue #413, opened 2017-08-30
Create enough sequence tests to make sure actions are created before sequences


daisy-ycguo:

mrutkows:
fixed in https://github.com/apache/incubator-openwhisk-wskdeploy/pull/483

PR #483, opened 2017-09-14
Add sequence validation tests

daisy-ycguo:
Close issue #413 

The test code will try 10 times to deploy sequences, composing with 2~11 actions, to verify if the package can be deployed successfully.

If an action is not created when sequence is creating, OpenWhisk will return errors which will cause the deployment fails.


daisy-ycguo:
@pritidesai @mrutkows please review.

pritidesai:
LGTM, thanks @daisy-ycguo

PR #493, opened 2017-09-15
Modify the error messages

### LINKED ISSUE #504: Error message for manifest file not found does not reflect -m (or -p) file names
If I run 
```
wskdeploy -i -m docs/examples/manifest_hello_world.yaml
```
and the yaml file is NOT found, I get the following error message today:
```
Error: /Users/Matt/go/src/github.com/apache/incubator-openwhisk-wskdeploy/cmd/root.go [235]: Invalid input of Yaml file =====> Missing manifest.yaml/manifest.yml file.
```

This message would be fine if I were using the 'path' flag ```-p``` as it SHOULD look for the default name of manifest.yaml (or .yml).  However, IF the user provides a manifest path (with actual filename, the error message should tell user something like "manifest file [ <value from -m flag> ] Not found at path [ <value from -p flag> ].

In this case, we should see something like:
```
Error: /Users/Matt/go/src/github.com/apache/incubator-openwhisk-wskdeploy/cmd/root.go [235]: Invalid input of Yaml file =====> Manifest file not found [manifest_hello_world.yaml] at path [ x ].
```
In fact the named Error should be **ErrorManifestFileNotFound**() with uniform output to repeat what the user provided on the command line for both the -m and -p flags (when provided).  

Of course, if the -p, flag were not provided then show the default directory where we looked (i.e.  assume this would display the same directory as the wskdeploy command).  If the -m was not provided, then of course we would display the default name 'manifest.yaml/manifest/yml' as we are doing now on the current error.

houshengbo:
Closes #504


daisy-ycguo:
LGTM.

PR #500, opened 2017-09-18
misc unitary test.go

lzbj:
Rename util_test.go to misc_test.go.


utils/misc_test.go
```diff
@@ -72,3 +72,23 @@ func TestDependencies(t *testing.T) {
 	assert.Equal(t, "http://github.com/user/repo", record.BaseRepo, "BaseRepo is wrong")
 	assert.Equal(t, "/subfolder1/subfolder2", record.SubFolder, "SubFolder is wrong")
 }
+
+func TestParseOpenWhisk(t *testing.T) {
+	openwhiskHost := "https://openwhisk.ng.bluemix.net"
```
mrutkows: Not sure how I feel about using this host/endpoint for Unit testing...

utils/misc.go
```diff
@@ -619,3 +619,42 @@ func GetDeploymentFilePath(projectPath string) string {
 	}
 }
 
+// agnostic util reader to fetch content from web or local path or potentially other places.
```
mrutkows: Did we get some optimization for moving these to misc.go since they are still part of same package?  Is this just reducing # of files we have for simplicity?

mrutkows:
@pritidesai I added a couple comments, but not a review... care to provide any comments?

Issue #492, opened 2017-09-15
Two "inputs" formats create different assets


daisy-ycguo:
According the samples in specification, we support two formats for "inputs" (list of
parameter):
a complex one:
```
  actions:
    func1:
      function: actions/function.js
      runtime: nodejs:6
      inputs:
        functionID:
          type: string
          description: the ID of function
        visited:
          type: string
          description: the visted city list
```
and a simple one:
```
package:
  name: TestSequencesCreation
  actions:
    func1:
      function: actions/function.js
      runtime: nodejs:6
      inputs:
        functionID: string
        visited: string
```
The complex one will create an action, with empty string `""` to be the default values of parameters. The simple one will create an action with `string` to be the default values of parameters. See below samples.
action created by complex format
```
{
    "namespace": "guoyingc@cn.ibm.com_dev/TestSequencesCreation",
    "name": "func1",
    "version": "0.0.1",
    "exec": {
        "kind": "nodejs:6",
        "code": "/**\n * Return a simple string to \n * confirm this function has been visited.\n *\n * @param visited the visited function list\n */\nfunction main(params) {\n    functionID = params.functionID || 'X'\n    if (params.visited == null) {\n        params.visited = 'function'+functionID;\n    } else {\n        params.visited = params.visited + ', function'+functionID;\n    }\n    return {\"visited\":params.visited};\n}\n"
    },
    "annotations": [
        {
            "key": "exec",
            "value": "nodejs:6"
        }
    ],
    "parameters": [
        {
            "key": "functionID",
            "value": ""
        },
        {
            "key": "visited",
            "value": ""
        }
    ],
    "limits": {
        "timeout": 60000,
        "memory": 256,
        "logs": 10
    },
    "publish": false
}
```
action created by the simple format:
```
ok: got action TestSequencesCreation/func1
{
    "namespace": "guoyingc@cn.ibm.com_dev/TestSequencesCreation",
    "name": "func1",
    "version": "0.0.1",
    "exec": {
        "kind": "nodejs:6",
        "code": "/**\n * Return a simple string to \n * confirm this function has been visited.\n *\n * @param visited the visited function list\n */\nfunction main(params) {\n    functionID = params.functionID || 'X'\n    if (params.visited == null) {\n        params.visited = 'function'+functionID;\n    } else {\n        params.visited = params.visited + ', function'+functionID;\n    }\n    return {\"visited\":params.visited};\n}\n"
    },
    "annotations": [
        {
[...]

PR #511, opened 2017-09-19
Fix single-line parameter to look for Type names (and default with proper value)

mrutkows:
Closes https://github.com/apache/incubator-openwhisk-wskdeploy/issues/492


mrutkows:
@daisy-ycguo there was a problem when I enabled all the unit tests (in manifest_parser_test.go)... so I commented out my code for now but will work on it tomorrow.  Please review what I did change and merge if OK (please accept the commented out code as I believe it is close to a fix in manifest_parser.go).

PR #513, opened 2017-09-19
Support string and env variable concatenation

### LINKED ISSUE #501: We don't support concatenation of an environment value and a string
We don't support environment using like this: $CLOUDANT_USERNAME.cloudant.com. `CLOUDANT_USERNAME.cloudant.com` will be treated as a whole part. And it will report an error "WARNING: Missing Environment Variable CLOUDANT_USERNAME.cloudant.com."

Do we need to support this feature ?

daisy-ycguo:
Close #501 

After this PR, we can support below format of values:
- $ENV_VAR
- ${ENV_VAR}
- yyyy${ENV_VAR}xxxx
- yyyy${ENV_VAR1}xxxx${ENV_VAR2}zzzz

### DIFF
--- a/utils/util_test.go
+++ b/utils/util_test.go
@@ -57,6 +57,13 @@ func TestGetEnvVar(t *testing.T) {
 	assert.Equal(t, "oh, dollars!", GetEnvVar("$WithDollar"), "dollar sign should be handled.")
 	assert.Equal(t, "5000", GetEnvVar("5000"), "Should be no difference between integer and string.")
 	assert.Equal(t, "", GetEnvVar("$WithDollarAgain"), "if not found in environemnt, return empty string.")
+	assert.Equal(t, "oh, dollars!.ccc.aaa", GetEnvVar("${WithDollar}.ccc.aaa"), "String concatenation fail")
+	assert.Equal(t, "ddd.NO dollar.aaa", GetEnvVar("ddd.${NoDollar}.aaa"), "String concatenation fail")
+	assert.Equal(t, "oh, dollars!.NO dollar.aaa", GetEnvVar("${WithDollar}.${NoDollar}.aaa"), "String concatenation fail")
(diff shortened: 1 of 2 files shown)


utils/util_test.go
```diff
@@ -57,6 +57,13 @@ func TestGetEnvVar(t *testing.T) {
 	assert.Equal(t, "oh, dollars!", GetEnvVar("$WithDollar"), "dollar sign should be handled.")
 	assert.Equal(t, "5000", GetEnvVar("5000"), "Should be no difference between integer and string.")
 	assert.Equal(t, "", GetEnvVar("$WithDollarAgain"), "if not found in environemnt, return empty string.")
+	assert.Equal(t, "oh, dollars!.ccc.aaa", GetEnvVar("${WithDollar}.ccc.aaa"), "String concatenation fail")
+	assert.Equal(t, "ddd.NO dollar.aaa", GetEnvVar("ddd.${NoDollar}.aaa"), "String concatenation fail")
+	assert.Equal(t, "oh, dollars!.NO dollar.aaa", GetEnvVar("${WithDollar}.${NoDollar}.aaa"), "String concatenation fail")
+	assert.Equal(t, "ddd.ccc.oh, dollars!", GetEnvVar("ddd.ccc.${WithDollar}"), "String concatenation fail")
+	assert.Equal(t, "", GetEnvVar("$WithDollarAgain.ccc.aaa"), "String concatenation fail")
+	assert.Equal(t, "ddd..aaa", GetEnvVar("ddd.${WithDollarAgain}.aaa"), "String concatenation fail")
+	assert.Equal(t, "oh, dollars!NO dollar.NO dollar", GetEnvVar("${WithDollar}${NoDollar}.${NoDollar}"), "String concatenation fail")
```
mrutkows: What happens with the following strings that have ($) dollar char and both parens...
1) $()
2) $abc()

PR #521, opened 2017-09-20
Updated .PHONY targets

GabrielNicolasAvellaneda:
All targets that do not represent file names should be phony targets


lzbj:
@GabrielNicolasAvellaneda , thanks for your PR.

PR #522, opened 2017-09-20
Check mandatory field Version and License

### LINKED ISSUE #507: 'version' key is mandatory for 'package' key only; optional for Actions
Previously, the spec. supported a version for all which entities.  However, I found via testing that we do not enforce any checks (beyond the YAML parser) on the 'version' key/field.

Because we have nothing enforced (and seeing how users are using this field), we have adjusted the spec. to match user expectations in v0.8.9 of the spec. as follows:

1) version is a mandatory field for 'package'
2) version is an optional key for 'action'
3) version is "not supported" or "invalid" for any other entity (e.g., Rule, Trigger) at this time

The version should adhere to the format as defined by Maven here

**_Maven-Version_** :
The version type is defined with the Apache Maven project’s policy   draft:   https://cwiki.apache.org/confluence/display/MAVEN/Version+number+policy

### LINKED ISSUE #508: 'license' key is mandatory for 'package' structure
Currently, it appears we do not support any checks/enforcement for the 'license' key for the 'package' structure.

It should be mandatory, and as the spec. indicates be one of the values referenced here:
Linux Foundation, SPDX License list: https://spdx.org/licenses/

If a valid license value is NOT found, we should issue a (new) named error such as **ErrorInvalidLicenseValue**().

We should discuss if this is too onerous for user to be forced to fill this in and explore setting the default to "unknown" and issuing a **WARNING** instead.

daisy-ycguo:
Closes #507 
Closes #508 

This PR will check the mandatory field "Version" and "License". If they are not exist, it will print two warning messages:
```
WARNING: Mandatory field Package Version must be set.
WARNING: Package Version is not saved in the current wskdeploy version.
```

This PR contains a validation check of a License. Firstly, it will check the license based on a local popular license list with a high performance. If the local check is failed, it will check based on a remote json file, with a lower performance.

### DIFF
--- a/parsers/manifest_parser.go
+++ b/parsers/manifest_parser.go
@@ -216,6 +216,31 @@ func (dm *YAMLParser) ComposePackage(pkg Package, packageName string, filePath s
 	pub := false
 	pag.Publish = &pub
 
+	//Version is a mandatory value
+	//If it is an empty string, it will be set to default value
+	//And print an warning message
+	if pkg.Version == "" {
+		warningString := wski18n.T("WARNING: Mandatory field Package Version must be set.\n")
+		whisk.Debug(whisk.DbgWarn, warningString)
+		warningString = wski18n.T("WARNING: Package Version is not saved in the current wskdeploy version.\n")
+		whisk.Debug(whisk.DbgWarn, warningString)
+		pkg.Version = "0.0.1"
+	}
+
+	//License is a mandatory value
+	//set license to unknown if it is an empty string
+	//And print an warning message
+	if pkg.License == "" {
+		warningString := wski18n.T("WARNING: Mandatory field Package License must be set.\n")
+		whisk.Debug(whisk.DbgWarn, warningString)
+		warningString = wski18n.T("WARNING: Package License is not saved in the current wskdeploy version.\n")
+		whisk.Debug(whisk.DbgWarn, warningString)
+		pkg.License = "unlicensed"
+	} else {
+		utils.CheckLicense(pkg.License)
+	}
+
+	//set parameters
 	keyValArr := make(whisk.KeyValueArr, 0)
 	for name, param := range pkg.Inputs {
 		var keyVal whisk.KeyValue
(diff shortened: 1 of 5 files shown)


wski18n/i18n_resources.go
```diff
@@ -15,449 +15,257 @@
...
-// wski18n/resources/zh_Hant.all.json
-// DO NOT EDIT!
-
 package wski18n
 
 import (
     "bytes"
     "compress/gzip"
     "fmt"
     "io"
```
houshengbo: I suspect this file is not correctly generated. Why are these imports removed?
Each time, this file is regenerated, only the some byte arrays change.

daisy-ycguo: I generated it based on your document wski18n/README.md.
```
go get -u github.com/jteeuwen/go-bindata/...
sudo apt-get install go-bindata
go-bindata -pkg wski18n -o wski18n/i18n_resources.go wski18n/resources;
```

parsers/manifest_parser.go
```diff
@@ -215,6 +215,29 @@ func (dm *YAMLParser) ComposePackage(pkg Package, packageName string, filePath s
...
+		whisk.Debug(whisk.DbgWarn, warningString)
+		warningString = wski18n.T("WARNING: Package Version is not saved in the current wskdeploy version.\n")
+		whisk.Debug(whisk.DbgWarn, warningString)
+		pkg.Version = "0.0.1"
+	}
+
+	//License is a mandatory value
+	//set license to unknown if it is an empty string
+	//And print an warning message
+	if pkg.License == "" {
```
mrutkows: Acccording to the specification, we need to verify if the value is:

•	A valid Package license value MUST be one of the Linux SPDX license values; for example: Apache-2.0 or GPL-2.0+, or the value ‘unlicensed’.

parsers/manifest_parser.go
```diff
@@ -215,6 +215,29 @@ func (dm *YAMLParser) ComposePackage(pkg Package, packageName string, filePath s
...
+		warningString := wski18n.T("WARNING: Mandatory field Package License must be set.\n")
+		whisk.Debug(whisk.DbgWarn, warningString)
+		warningString = wski18n.T("WARNING: Package License is not saved in the current wskdeploy version.\n")
+		whisk.Debug(whisk.DbgWarn, warningString)
+		pkg.License = "unknown"
```
mrutkows: default should be 'unlicensed'; this was my mistake, should have referenced the spec. first before commenting on the issue.

PR #532, opened 2017-09-21
Implement the trigger deletion in the correct order

houshengbo:
So far we delete the trigger and then invoke the feed, which leads
to error for sure, because we need to invoke the feed and then
delete the trigger. This PR implements the trigger deletion in
the correct order.


pritidesai:
LGTM

PR #534, opened 2017-09-21
Remove the trigger if feed fails to be called

houshengbo:
If feed fails to be called, we need to remove the trigger to keep
the environment clean.


pritidesai:
LGTM

PR #541, opened 2017-09-22
Add strict flag for user defined runtime.

lzbj:


parsers/manifest_parser.go
```diff
@@ -389,14 +389,14 @@ func (dm *YAMLParser) ComposeActions(filePath string, actions map[string]Action,
...
-			}
-		} else {
-			errStr := wski18n.T("wskdeploy has chosen a particular runtime for the action.\n")
-			whisk.Debug(whisk.DbgWarn, errStr)
-		}
+
+			} else if utils.Flags.Strict {
+                wskaction.Exec.Kind = action.Runtime
+            } else {
+                errStr := wski18n.T("wskdeploy has chosen a particular runtime for the action.\n")
```
houshengbo: This message should be added into json file as the i18n string. After that, i18n_resources should be regenerated.

lzbj: @houshengbo , thanks for your review, I think this message has already been there, and corresponding resource file are regenerated.
https://github.com/apache/incubator-openwhisk-wskdeploy/blob/master/wski18n/resources/en_US.all.json#L55

mrutkows:
Added new issue https://github.com/apache/incubator-openwhisk-wskdeploy/issues/553 to request test coverage.

PR #542, opened 2017-09-22
Fix an issue of deploymentreader and add a unit test

daisy-ycguo:
This PR is to
- Fix issues of deploymentreader.go
- Add a unit test

More integration tests will be added to test deploymentreader.go in another PR.


pritidesai:
LGTM

PR #555, opened 2017-09-26
Update Programming Guide - part7

mrutkows:
Added more Action examples using a new std. template for each example.


mrutkows:
@daisy-ycguo please review and merge if able.

PR #580, opened 2017-10-02
Add the support of the key 'packages' in deployment.yaml

### LINKED ISSUE #572: Deployment file should not require "application" key if manifest does not
This deployment file throws an error:
```
  packages:
      hello_world_package:
        triggers:
          meetPerson:
            inputs:
              name: Elrond
              place: Rivendell
              children: 3
              height: 1.88
```

this does not:
```
application:
  packages:
      hello_world_package:
        triggers:
          meetPerson:
            inputs:
              name: Elrond
              place: Rivendell
              children: 3
              height: 1.88
```
manifests can have their outermost document key be either "package" or "packages"; this should also be allowed for corresponding Deployment files.

Additionally,  if the deployment file DOES contain an application with a "name"  AND the manifest also contains an application "name" we should ERROR if there is a mismatch.

houshengbo:
Closes #572


parsers/yamlparser.go
```diff
@@ -178,14 +178,16 @@ type Application struct {
 }
 
 type DeploymentYAML struct {
-	Application Application `yaml:"application"` //used in deployment.yaml
+	Application Application       `yaml:"application"` //used in deployment.yaml
+    Packages   map[string]Package `yaml:"packages"` //used in deployment.yaml
```
mrutkows: Unfortunately, we should support "Package" (singular) here as well if we want to cover all combinations (even though we see how deploymentreader.go is getting a lot of if-else logic to allow for all these combinations).

deployers/deploymentreader.go
```diff
@@ -65,8 +65,17 @@ func (reader *DeploymentReader) bindPackageInputsAndAnnotations() {
 
 	if reader.DeploymentDescriptor.Application.Packages == nil {
 		// a single package is specified in deployment YAML file with "package" key
-		packMap[reader.DeploymentDescriptor.Application.Package.Packagename] = reader.DeploymentDescriptor.Application.Package
-        utils.PrintOpenWhiskOutputln("WARNING: The package YAML key in deployment file will soon be deprecated. Please use packages instead as described in specifications.")
+        if len(reader.DeploymentDescriptor.Application.Package.Packagename) != 0 {
```
mrutkows: Can we better comment and align the if-else blocks?

wski18n/resources/en_US.all.json
```diff
@@ -330,5 +330,10 @@
   {
     "id": "path of the .cert file",
     "translation": "path of the .cert file"
+  },
+  {
+    "id": "The name of the application {{.appNameDeploy}} in deployment.yaml does not match the name of the application {{.appNameManifest}}} in manifest.yaml.",
```
mrutkows: Although our tests assume the file names are manifest.yaml and deployment.yaml... the user could provide any name for the base file name.

houshengbo: Done.

deployers/deploymentreader.go
houshengbo: Done.

parsers/yamlparser.go
houshengbo: done

mrutkows:
@houshengbo @pritidesai notice now how we now have 2 structures that are exactly the same now in yamlparser.go..

```
type DeploymentYAML struct {
	Application Application       `yaml:"application"` //used in deployment.yaml
    Packages   map[string]Package `yaml:"packages"` //used in deployment.yaml
    Package    Package            `yaml:"package"`
	Filepath    string      //file path of the yaml file
}

type ManifestYAML struct {
	Package  Package `yaml:"package"` //used in both manifest.yaml and deployment.yaml
	Packages map[string]Package `yaml:"packages"`
	Filepath string  //file path of the yaml file
    Application Application       `yaml:"application"` //used in manifest.yaml
}
```

time for a simplification? or is there value (future?) keeping them separated?

PR #587, opened 2017-10-03
Create the basic struct for Deployment and Manifest yaml

houshengbo:


houshengbo:
@mrutkows I have to get rid of deployment and manifest yamls, and replace them with one yaml struct. The inheritance does not work if it is used to parse yaml files.

PR #596, opened 2017-10-05
publish to docker

beemarie:


beemarie:
Closes #597

.travis.yml
```diff
@@ -22,6 +22,7 @@ after_success:
 - DEPLOY_BUILD_READY=true
 after_script:
 - make clean
+- ./tools/travis/docker/publish.sh apache wskdeploy ${TRAVIS_TAG}
```
houshengbo: We need to move this line into a new section of provider.

deploy:
-provider: releases
 ...
 ...
- provider: script
  script: "./tools/travis/docker/publish.sh apache wskdeploy ${TRAVIS_TAG}"
  on:
    tags: true
    repo: apache/incubator-openwhisk-wskdeploy
    condition: "$DEPLOY_BUILD_READY = true"

tools/travis/docker/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM openwhisk/nodejs6action
```
houshengbo: Do we only expect nodejs?

beemarie: yes, we're building on top of this dockerfile to run the package action.

.travis.yml
beemarie: thanks -- this wasn't originally working, until I realized i needed skip_cleanup = true.  Should be good to go now.

PR #601, opened 2017-10-09
Add Json support for Input parameters

mrutkows:


pritidesai:
Its printing param details like:

```
./wskdeploy -m tests/dat/manifest_validate_json_params.yaml 
BEFORE:
        member2: (*parsers.Parameter)
                Parameter.Descrption: []
                Parameter.Type: []
                Parameter.Value: [map[name:Sam place:Shire]]
                Parameter.Default: [<nil>]
EXIT: Parameter type=[json] value=[map[name:Sam place:Shire]]
BEFORE:
...
```

mrutkows:
@pritidesai that's debug trace I added early July, it will be commented out before WIP is removed.

Issue #605, opened 2017-10-13
How do we prevent Concurrent Access to same OpenWhisk entity?


pritidesai:
We have seen couple of concurrency issues with deployment/undeployment when

- triggers and rules

When we try to run `wskdeploy` to with these two manifest files in deployment mode, deployment fails with `Error: Error: /Users/pritidesai/Documents/goworkspace/src/github.com/apache/incubator-openwhisk-wskdeploy/deployers/servicedeployer.go [913]: OpenWhisk Client Error =====> The requested resource does not exist. (code 23582280) Error code: 148.`

package1-manifest.yaml

```
packages:
  pacakge1:
      actions:
        helloworld:
          function: actions/helloworld.js
          runtime: nodejs:6
      triggers:
        locationupdate:
      rules:
        myRule:
          trigger: locationupdate
          rule: myRule
``` 

package2-manifest.yaml

```
packages:
  pacakge2:
      actions:
        helloworld:
          function: actions/helloworld.js
          runtime: nodejs:6
      triggers:
        locationupdate:
      rules:
        myRule:
          trigger: locationupdate
          rule: myRule
``` 

- dependency

Its a valid use case of having dependency on same package from different manifest files but this leads to the same issue of concurrency and sometimes fails with the same error 148 and sometimes fails with `"Error: /Users/pritidesai/Documents/goworkspace/src/github.com/apache/incubator-openwhisk-wskdeploy/deployers/servicedeployer.go [666]: OpenWhisk Client Error =====> Sequence component does not exist. (code 23582010) Error code: 144.\n"})` or `Deployment of dependency myhelloworlds did not complete sucessfully. Run `wskdeploy undeploy` to remove partially deployed assets.\nError: /Users/pritidesai/Documents/goworkspace/src/github.com/apache/incubator-openwhisk-wskdeploy/deployers/servicedeployer.go [666]: OpenWhisk Client Error =====> Concurrent modification to resource detected. (code 23549814) Error code: 153.\n"})`

package1-manifest.yaml

```
packages:
  package1:
      dependencies:
          myhelloworlds:
                location: github.com/apache/incubator-openwhisk-test/packages/helloworlds
      actions:
        hello:
          function: actions/helloworld.js
          runtime: nodejs:6
       sequences:
          helloworldnodejs-series:
                actions: hello, myhelloworlds/hello-js
```

package2-manifest.yaml

```
packages:
  package2:
      dependencies:
          myhelloworlds:
[...]

PR #606, opened 2017-10-13
Add the dependency back to validate packages

houshengbo:


mrutkows:
@houshengbo I would like @pritidesai to be able to review and comment as she has been identifying and addressing issues where we are having namespace collisions and seeing errors which caused us to disable this originally.

pritidesai:
These changes can lead us to build failure on master branch or any of the PRs, I have seen failures in my local environment due to concurrency issue. I suggest not merging this PR until we fix #605.

mrutkows:
@houshengbo will re-open once we have the changes from 605.

PR #607, opened 2017-10-16
JSON Output support 

mrutkows:
First part of PR we are "Updating Bad YAML tests to be more descriptive and efficient."


parsers/manifest_parser_test.go
```diff
@@ -339,15 +324,15 @@ func TestParseManifestForMultiLineParams(t *testing.T) {
...
-        if payload, ok := action.Outputs["payload"]; ok {
-            p := payload.(map[interface{}]interface{})
-            expectedResult = "string"
-            actualResult = p["type"].(string)
-            assert.Equal(t, expectedResult, actualResult, "Expected " + expectedResult + " but got " + actualResult)
-            expectedResult = "parameter dump"
-            actualResult = p["description"].(string)
-            assert.Equal(t, expectedResult, actualResult, "Expected " + expectedResult + " but got " + actualResult)
-        }
+        //if payload, ok := action.Outputs["payload"]; ok {
```
houshengbo: Why do you comment out these lines in this test?

mrutkows: @houshengbo  I was wondering if you would catch this (saw this just before I asked you to review) ... i had removed it because:
1) Outputs are now truly of type Parameter (and original code would not work and for got about it).
2) I believed we needed more dedicated/comprehensive output tests for single, multi and JSON parameter types and had thought to open a new issue to address (which I just did):

Will put back the previous payload assertion for now...

and also opened this issue: https://github.com/apache/incubator-openwhisk-wskdeploy/issues/620

Issue #610, opened 2017-10-19
Provide support for an env_variable for package name


beemarie:
Currently in the deploy package, if we want to provide a different package name, we are parsing the file & renaming the package name node.  Would it be possible to provide support for an optional environment variable as the package name?

csantanapr:
@mrutkows @houshengbo @daisy-ycguo 
The sample is the following, we want to be able to have a manifest of this form
```
package:
  name: $MY_AWESOME_PACKAGE
  version: 1.0
  license: Apache-2.0
  namespace: _
  actions:
    helloworld:
      function: actions/helloworld.js
```
And  use the environment variable `MY_AWESOME_PACKAGE`  for the package name.

@beemarie what's the current behavior? you get an error in wskdeploy?

houshengbo:
@beemarie @csantanapr 
We are deprecating the format in the previous comment. Here is the new format looks like:
```
packages:
  PACKAGE_NAME:
    version: 1.0
    license: Apache-2.0
    namespace: _
    actions:
      helloworld:
        function: actions/helloworld.js
```
Do we need the package name able to read from env variable supported for the new format?
Besides, is it only for package name? We also have trigger name, dependency name, rule name, etc.

I have to say that if we support env variables to be picked up by the manifest, it can be even more complicated to maintain the yaml files, because there can be multiple env variables to manage.

beemarie:
@houshengbo @csantanapr 
It is more important to be able to provide PACKAGE_NAME as an env-variable over trigger name, rule name, etc., as packages can not be renamed once created.

mrutkows:
it could be implemented as a new feature easily with exiting code; however, we are moving away from “package” to “packages” and then it becomes more difficult...

Currently, "packages" follows our consistent convention/pattern to use names as "keys" (or IDs) in the YAML document. The YAML parser may not handle having and odd string (it might though would have to try it with the go-yaml parser).  If the parser cannot handle it, then we would have to assure we had the “name” (or similar) field as a valid field for “packages” and make sure we replaced the name on the API call.

mrutkows:
@houshengbo also be aware that we allow "partial" replacement as well...
e.g.
```
packages:
  IBM-{$PACKAGE_NAME}:
    version: 1.0
    license: Apache-2.0
    namespace: _
    actions:
      helloworld:
        function: actions/helloworld.js
```

mrutkows:
[...]

PR #614, opened 2017-10-20
Add the support to read env variable for package name

houshengbo:
Closes: #610

### DIFF
--- a/parsers/deploy_parser.go
+++ b/parsers/deploy_parser.go
@@ -53,7 +53,8 @@ func (dm *YAMLParser) ParseDeployment(deploymentPath string) (*YAML, error) {
         return &dplyyaml, utils.NewParserErr(deploymentPath, lines, msgs)
     }
 	dplyyaml.Filepath = deploymentPath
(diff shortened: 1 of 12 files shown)


pritidesai:
@houshengbo please add a sample manifest files with single and multiple packages:

```
packages:
   $package_name:
      actions:
         sample-action:
            function: sample-action.js
            inputs:
               param: value
```


```
packages:
   $package_name_1:
      actions:
         sample-action:
            function: sample-action.js
            inputs:
               param: value
   $package_name_2:
      actions:
         sample-action:
            function: sample-action.js
            inputs:
               param: value
```

mrutkows:
@houshengbo thanks for adding additional tests for the "concatenation" of env. variable strings with in-line strings.

mrutkows:
@houshengbo Priti brings up a validate request for an additional test(s) to assure it works on multiples (which it should looking at the code).

houshengbo:
@pritidesai @mrutkows I have added the coverage of multiple packages.

PR #628, opened 2017-10-26
WIP: Adding action annotations while creating that action

### LINKED ISSUE #609: Annotations on Actions only work if "web export" exists
It appears at some point (perhaps when we added support for "web actions" that the ComposeActions() function has logic that ONLY creates Annotations on the wskAction IFF ```action.Webexport == "true"```. 

To make matters more convoluted, even if "Webexport==true", the annotations are created as a "side effect" of calling ```utils.WebAction```; this is not good coding practice.

```		keyValArr = make(whisk.KeyValueArr, 0)
		for name, value := range action.Annotations {
			var keyVal whisk.KeyValue
			keyVal.Key = name
			keyVal.Value = utils.GetEnvVar(value)
			keyValArr = append(keyValArr, keyVal)
		}

		// only set the webaction when the annotations are not empty.
		if action.Webexport == "true" {
			// TODO() why is this commented out?  we should now support annotations...
			//wskaction.Annotations = keyValArr
			wskaction.Annotations, errorParser = utils.WebAction("yes", keyValArr, action.Name, false)
			if errorParser != nil {
				return s1, errorParser
			}
		}
```

we have NO testcases for Annotations on Actions (only for Triggers) and need several as part of fixing this issue.  The Annotations SHOULD work with or without "webExport==true" and NOT be a side effect.

pritidesai:
Closes #609


pritidesai:
Created new PR based on the latest master #631

PR #632, opened 2017-10-27
Add unit tests for wskdeploy custom error types.

mrutkows:


mrutkows:
@pritidesai @houshengbo please review when able.

pritidesai:
Thanks @mrutkows LGTM with couple of questions:

I tried running it with a simple YAML error (commenting out `packages` key from manifest file) which printed the manifest file name twice, which I found little strange:

```
$ ./wskdeploy -m tests/src/integration/managed-deployment/00-manifest-minus-second-package.yaml  --managed
Error: manifestreader.go [49]: [ERROR_FILE_READ_ERROR]: File: [00-manifest-minus-second-package.yaml]: manifest_parser.go [96]: [ERROR_YAML_PARSER_ERROR]: File: [00-manifest-minus-second-package.yaml]: ====> yaml: line 3: did not find expected key
```

mrutkows:
@pritidesai Thanks!
I will look at that one... it may be an inheritance problem in the error messages.

mrutkows:
@pritidesai actually, it may be a newline character on YAMLParserError

mrutkows:
@pritidesai actually, its performing as expected given a new error is being created with the message from another error...
```
	manifest, err := manifestParser.ParseManifest(dep.ManifestPath)
	if err != nil {
		return manifest, manifestParser, utils.NewFileReadError(dep.ManifestPath, err.Error())
	}
```

mrutkows:
So it looks like the code should instead of creating a new (less informative) error just surface the lower-level error...
```
manifest, err := manifestParser.ParseManifest(dep.ManifestPath)

//	if err != nil {
//		return manifest, manifestParser, utils.NewFileReadError(dep.ManifestPath, err.Error())
//	}
	return manifest, manifestParser, err  // instead of nil
```

mrutkows:
@pritidesai however looking in the same file... (manifestreader.go), it seems we attempt to do this same creating of a new (less informative) error when a lower level error has been detected...

For example,
```
func (reader *ManifestReader) InitRootPackage(manifestParser *parsers.YAMLParser, manifest *parsers.YAML, ma whisk.KeyValue) error {
	packages, err := manifestParser.ComposeAllPackages(manifest, reader.serviceDeployer.ManifestPath, ma)
	if err != nil {
		return utils.NewYAMLFileFormatError(manifest.Filepath, err.Error())
	}
	reader.SetPackage(packages)

	return nil
}
```
and about a dozen times in...
```
unc (deployer *ManifestReader) HandleYaml(sdeployer *ServiceDeployer, manifestParser *parsers.YAMLParser, manifest *parsers.YAML, ma whisk.KeyValue) error {

	var err error
	var manifestName = manifest.Filepath

	deps, err := manifestParser.ComposeDependenciesFromAllPackages(manifest, deployer.serviceDeployer.ProjectPath, deployer.serviceDeployer.ManifestPath)
	if err != nil {
		return utils.NewYAMLFileFormatError(manifestName, err.Error())
	}

	actions, err := manifestParser.ComposeActionsFromAllPackages(manifest, deployer.serviceDeployer.ManifestPath, ma)
	if err != nil {
		return utils.NewYAMLFileFormatError(manifestName, err.Error())
	}
```
we may need another issue to clean this up... and it may mean that the "YAMLFileFormatError" may become less used and perhaps obsolete.

mrutkows:
@pritidesai @houshengbo opened https://github.com/apache/incubator-openwhisk-wskdeploy/issues/648

mrutkows:
@pritidesai addressed my own "TODO" and tested YAML parser error line numbers and error messages (returned by the YAMLParser library).   Also, updated the logic that generates these indented messages to be cleaner.

pritidesai:
Looks much more cleaner now, thanks for adding this @mrutkows 

LGTM

PR #661, opened 2017-11-30
Remove unnecessary rule changes

houshengbo:
A rule is enabled when it is created. There is no need to activate it.
A rule is able to delete no matter it is active or inactive. This PR
removes the unnecessary rule changes when creating and deleting a rule.

In addition, the order to undeploy assets needs to be changed to avoid
concurrency issue as well. All the entity creation and deletion need to
be retried if conflict is returned.


houshengbo:
@mrutkows 
I have run make integration_test 40 times in a row, and all the results turned out fine.
Not sure about 100 times or even 400 times, but this can be used as the fix in wskdeploy.
Regarding the retry in client go, I will figure out a way to add the retry as a config option.

Issue #651, opened 2017-11-28
Concurrency error: Error code: 153: Concurrent modification to resource detected. (code 379)


mrutkows:
Travis:

```
==> Executing: wskdeploy undeploy -m /home/travis/gopath/src/github.com/apache/incubator-openwhisk-wskdeploy/tests/src/integration/validate-manifest-deployment-file-extensions/not-standard-manifest.yaml -d /home/travis/gopath/src/github.com/apache/incubator-openwhisk-wskdeploy/tests/src/integration/validate-manifest-deployment-file-extensions/not-standard-deployment.yaml

==> Error: wskdeploy.go [125]: [ERROR_COMMAND_FAILED]: Command: [wskdeploy]: Error: servicedeployer.go [1169]: [ERROR_WHISK_CLIENT_ERROR]: Error code: 153: Concurrent modification to resource detected. (code 379).

--- FAIL: TestNonStandardFileNames (0.65s)

	Error Trace:	validate-file-extensions_test.go:58
	Error:      	Not equal: 
	            	expected: <nil>(<nil>)
	            	actual: *utils.CommandError(&utils.CommandError{WskDeployBaseErr:utils.WskDeployBaseErr{ErrorType:"ERROR_COMMAND_FAILED", FileName:"wskdeploy.go", LineNum:125, Message:"Command: [wskdeploy]: Error: servicedeployer.go [1169]: [ERROR_WHISK_CLIENT_ERROR]: Error code: 153: Concurrent modification to resource detected. (code 379)\n"}, Command:"wskdeploy"})

	Messages:   	Failed to undeploy based on the manifest and deployment files with non standard names.
```

mrutkows:
and now on a different test...

```
=== RUN   TestSupportProjectPath

==> Executing: wskdeploy -p /home/travis/gopath/src/github.com/apache/incubator-openwhisk-wskdeploy/tests/src/integration/flagstests

==> Output: WARNING: The authentication key was retrieved from whisk.properties which will soon be deprecated please do not use it outside of Travis builds.

WARNING: The API host was retrieved from whisk.properties which will soon be deprecated please do not use it outside of Travis builds.

Deployment completed successfully.

.

==> Executing: wskdeploy undeploy -p /home/travis/gopath/src/github.com/apache/incubator-openwhisk-wskdeploy/tests/src/integration/flagstests

[...]

PR #663, opened 2017-12-01
Make travis faster

csantanapr:
This makes deployment Travis faster
Also address some tests errors when controllers are in HA giving back 409s conflict errors on entities


csantanapr:
Closes #651

PR #665, opened 2017-12-01
Refactor "error" related data/functions into their own pkg to avoid cycles.

mrutkows:


mrutkows:
@houshengbo please review this refactoring, thx!

PR #666, opened 2017-12-07
Uncommenting managed deployment for rules, package, actions, and sequences

pritidesai:
After spending an hour the night before a big showcase on Tuesday, found these few lines commented out which was causing managed deployment to fail :(


houshengbo:
@pritidesai Do we have test cases indicating the failures? How did you find the failures?

pritidesai:
@houshengbo we have test cases to validate managed deployment but it confirms that deployment succeeds but there is no way to validate whether the entities were deleted on the server when they were on client.

houshengbo:
All the failures are related to 'swift:3'.

PR #675, opened 2017-12-20
Adding message hub test back

pritidesai:
Adding integration test for message hub back. Also, added response body and status in WhiskClientError.

Integration tests were mainly failing with the trigger "wskdeployMessageHubTrigger". Integration test was successful after renaming this trigger to something else. I couldn't determine the exact cause of failure. But it was weird to see that this trigger was not part of trigger listing even after trigger was created. Also, when the POST command was run manually, it was failing with "no permission to access this resource".  There might be some conflict with some other trigger or trigger ID on OpenWhisk side as its under shared account.

Adding HTTP response info into WhiskClientError as:

```
Error: servicedeployer.go [846]: [ERROR_WHISK_CLIENT_ERROR]: Error code: 246: The following application error was received: {}: HTTP Response Status: application error HTTP Response Body: {"duration":916,"name":"messageHubFeed","subject":"*","activationId":"0cd2c08003684dd892c0800368edd888","publish":false,"annotations":[{"key":"path","value":"whisk.system/messaging/messageHubFeed"},{"key":"waitTime","value":90},{"key":"kind","value":"nodejs:6"},{"key":"limits","value":{"timeout":60000,"memory":256,"logs":10}},{"key":"initTime","value":385}],"version":"0.0.35","response":{"result":{"error":"You must supply a 'topic' parameter."},"success":false,"status":"application error"},"end":1513732045616,"logs":[],"start":1513732044700,"namespace":"*"}
```

Compared to right now:

```
Error: servicedeployer.go [846]: [ERROR_WHISK_CLIENT_ERROR]: Error code: 246: The following application error was received: {}
```

### DIFF
--- a/deployers/servicedeployer.go
+++ b/deployers/servicedeployer.go
@@ -22,24 +22,25 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
-	"reflect"
 	"time"
 
 	"github.com/apache/incubator-openwhisk-client-go/whisk"
 	"github.com/apache/incubator-openwhisk-wskdeploy/parsers"
 	"github.com/apache/incubator-openwhisk-wskdeploy/utils"
-	"github.com/apache/incubator-openwhisk-wskdeploy/wski18n"
 	"github.com/apache/incubator-openwhisk-wskdeploy/wskderrors"
+	"github.com/apache/incubator-openwhisk-wskdeploy/wski18n"
+	"net/http"
 )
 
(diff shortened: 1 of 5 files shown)


mrutkows:
Nice investigation Priti... naming collisions will only grow over time; at some point soon, we should at least revisit allowing triggers and rules to be namespaced. If I recall, most were in favor, but would require the server-side work to support.

mrutkows:
@pritidesai and thanks for adding the response data to the errors for all the client calls.  That will be really very helpful.

PR #679, opened 2018-01-04
Use proper IDs when referencing i18n strings. (Base changes + phase 1 conversions)

mrutkows:
This PR lays the groundwork for the change and includes about 1/3 of the change over to using IDs. The plan is to follow-on the other 2/3 changes in future PRs.


mrutkows:
This PR lays the groundwork for the change and includes about 1/3 of the change over to using IDs.  The plan is to follow-on the other 2/3 changes in future PRs.

mrutkows:
So far, this exercise has uncovered missing translations, translations not used, and ways to simplify the # of duplicate/similar messsages.

PR #732, opened 2018-02-13
adding env. variable support in trigger feed

pritidesai:


parsers/manifest_parser.go
```diff
@@ -774,6 +774,10 @@ func (dm *YAMLParser) ComposeTriggers(filePath string, pkg Package, ma whisk.Key
 			trigger.Feed = trigger.Source
 		}
 
+		// replacing env. variables here in the trigger feed name
```
mrutkows: Glad you found this and had a ready-made fx to do the interpolation.

parsers/manifest_parser_test.go
```diff
@@ -1280,6 +1280,9 @@ func TestComposeSequences(t *testing.T) {
 }
 
 func TestComposeTriggers(t *testing.T) {
+	// set env variables needed for the trigger feed
+	os.Setenv("KAFKA_INSTANCE", "kafka-broker")
```
mrutkows: Kafka is a good choice for unit test.

tests/dat/manifest_data_compose_triggers.yaml
```diff
@@ -26,3 +26,8 @@ package:
       inputs:
         name: myname
         place: myplace
+    message-trigger:
+      feed: Bluemix_${KAFKA_INSTANCE}_Credentials-1/messageHubFeed
```
mrutkows: excellent!

PR #733, opened 2018-02-13
Enable Apache Rat plugin in gradle

daisy-ycguo:
After applying this PR, we can use below command to check headers:
`gradle rat`


pritidesai:
LGTM, thanks @daisy-ycguo

PR #738, opened 2018-02-14
GH-735 adjusting check for github to mainly make sure that github app…

### LINKED ISSUE #735: Add support for credentials for private git repo dependencies
Dependencies allows you to define a location and version of the artifact that you want to use. Here is an example:

```
project:
  name: MyProject
  version: 0.0.1
  packages:
    mypackage:
      license: Apache-2.0
      version: 0.0.1
      dependencies:
        mydep:
          location: https://github.com/my-org/my-private-repo
          version: 0.0.1
```

However, there does not appear to be a way to define credentials for accessing this repository - which is required for private repos.

The error you get when you try to do this is also not great (it is calling: https://github.com/my-org/my-private-repo/zipball/0.0.1):

```
Error: manifestreader.go [103]: [ERROR_YAML_FILE_FORMAT_ERROR]: File: [manifest.yaml]:
==> manifestreader.go [147]: [ERROR_YAML_FILE_FORMAT_ERROR]: File: [mypackage]:
==> ==> zip: not a valid zip file
==>
```

What is actually happening is the call results in a 404 from github.

allen-servedio:
…ears in the host name. Will allow many variations of host names as well as username and password.

Closes #735

### DIFF
--- a/utils/dependencies_test.go
+++ b/utils/dependencies_test.go
@@ -25,6 +25,10 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+/* NOTE: Though username and password is used here, this does not mean that wskdeploy fully supports private repos.
+ * This is merely one of many changes required to support them.
+ */
+
 func TestLocationIsGithub(t *testing.T) {
 	assert.True(t, LocationIsGithub("github.com/my-org/my-project"), "Does not allow github without a http/https prefix")
 	assert.True(t, LocationIsGithub("github.ibm.com/my-org/my-project"), "Does not allow github.ibm.com without a http/https prefix")
--- a/utils/dependencies_test.go
+++ b/utils/dependencies_test.go
@@ -1,0 +1,55 @@
+// +build unit
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package utils
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


utils/dependencies_test.go
```diff
@@ -0,0 +1,55 @@
...
+package utils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLocationIsGithub(t *testing.T) {
+	assert.True(t, LocationIsGithub("github.com/my-org/my-project"), "Does not allow github without a http/https prefix")
```
mrutkows: Why do all the "True" assertions (that test the existence of the protocol prefix) have a message string that indicates it expects a failure (i.e., "Does not allow...")?

utils/dependencies_test.go
```diff
@@ -0,0 +1,55 @@
...
+	assert.True(t, LocationIsGithub("github.ibm.com/my-org/my-project"), "Does not allow github.ibm.com without a http/https prefix")
+	assert.True(t, LocationIsGithub("http://github.com/my-org/my-project"), "Does not allow github with a http/https prefix")
+	assert.True(t, LocationIsGithub("http://github.ibm.com/my-org/my-project"), "Does not allow github.ibm.com with a http/https prefix")
+}
+
+func TestLocationIsGithub_WithUsernamePassword(t *testing.T) {
+	assert.True(t, LocationIsGithub("username:password@github.com/my-org/my-project"), "Does not allow username/password and github without a http/https prefix")
```
mrutkows: This is new functionality... the working assumption for wskdeploy was it would work with public repos. only.  So this PR is a feature request.

mrutkows: Perhaps we should return more info from the function like the uname:pwd and flag indicating public/private on the result set for the caller to use.

allen-servedio: You can see I took the path of least resistance with this PR. I am happy to hear what a proper solution is from you all and adjust this PR as needed.

pritidesai: Thanks @allen-servedio for initiating this effort. Your PR looks great except that these changes creates perception that `wskdeploy` supports private GitHub repos, which is not true.  Many times, people look for examples and just `grep` for specific keywords like `username:password@github.com. Can you please add a comment in the test file saying that we still do not have support for private repo but this is one of many changes for adding that support and we can get rid of such comment once its supported. Thoughts?

allen-servedio: Will do.

mrutkows:
@allen-servedio Please remind me have you submitted you Apache ICLA?

allen-servedio:
@mrutkows - yes it was submitted and received by @csantanapr last Friday.

mrutkows:
@allen-servedio you were too fast ... i looked it up in the Apache directory and removed my comment )

PR #754, opened 2018-02-24
Adding support for default package

### LINKED ISSUE #681: proper handling of 'default' package name
'default' is a reserved package name. wskdeploy should use this name to deploy actions in the _package without name_ (e.g /ns/actionname)

pritidesai:
Closes #681


deployers/servicedeployer.go
```diff
@@ -703,9 +701,16 @@ func (deployer *ServiceDeployer) RefreshManagedPackages(ma map[string]interface{
...
 	for _, pack := range deployer.Deployment.Packages {
-		err := deployer.createPackage(pack.Package)
-		if err != nil {
-			return err
+		// "default" package is a reserved package name
+		// all openwhisk entities will be deployed under
+		// /<namespace> instead of /<namespace>/<package> and
+		// therefore skip creating a new package and set
+		// deployer.DeployActionInPackage to false which is set to true by default
+		if strings.ToLower(pack.Package.Name) != parsers.DEFAULT_PACKAGE {
```
mrutkows: Thanks for creating a constant.

deployers/servicedeployer.go
```diff
@@ -945,7 +946,7 @@ func (deployer *ServiceDeployer) createRule(rule *whisk.Rule) error {
 // Utility function to call go-whisk framework to make action
 func (deployer *ServiceDeployer) createAction(pkgname string, action *whisk.Action) error {
 	// call ActionService through the Client
-	if deployer.DeployActionInPackage {
+	if strings.ToLower(pkgname) != parsers.DEFAULT_PACKAGE {
```
mrutkows: normalized

deployers/servicedeployer.go
```diff
@@ -1136,9 +1137,15 @@ func (deployer *ServiceDeployer) UnDeployDependencies() error {
...
 func (deployer *ServiceDeployer) UnDeployPackages(deployment *DeploymentProject) error {
 	for _, pack := range deployment.Packages {
-		err := deployer.deletePackage(pack.Package)
-		if err != nil {
-			return err
+		// "default" package is a reserved package name
+		// all openwhisk entities were deployed under
+		// /<namespace> instead of /<namespace>/<package> and
+		// therefore skip deleting default package during undeployment
+		if strings.ToLower(pack.Package.Name) != parsers.DEFAULT_PACKAGE {
```
mrutkows: Do we still need to delete it from /namespace ?

mrutkows: What about update/sync?

parsers/manifest_parser.go
```diff
@@ -322,6 +323,13 @@ func (dm *YAMLParser) ComposePackage(pkg Package, packageName string, filePath s
 		pag.Annotations = append(pag.Annotations, ma)
 	}
 
+	// "default" package is a reserved package name
+	// and in this case wskdeploy deploys openwhisk entities under
+	// /namespace instead of /namespace/package
+	if strings.ToLower(pag.Name) == DEFAULT_PACKAGE {
+		wskprint.PrintlnOpenWhiskInfo(wski18n.T(wski18n.ID_MSG_DEFAULT_PACKAGE))
```
mrutkows: tyvm

deployers/servicedeployer.go
pritidesai: Nope, we do not have to delete it from /namespace as its a reserved package. Its can neither be created nor be deleted. Its pretty much same as having no package and therefore the entities are created directly under `/namespace` except in case of web actions where action URL has `default` specified in it.

pritidesai:
thank you @mrutkows for the review, these changes have been tested against managed deployment as well. Here the high level flow is when a manifest file has a package name set to `default`, package creation is skipped, action is created directly under namespace. While creating rule, action is referred by `/namespace/action` instead of /namespace/package/action`. Also, the flag `DeployActionInPackage` is deleted here as this flag was only utilized during deployment and could not be used during undeployments. It was dropped to be consistent with a way we check if entities are under namespace or part of a non default package.

PR #776, opened 2018-03-08
Adding support for Inline Code and reading action source from URL

### LINKED ISSUE #622: Add unit tests to support "in-line" code on "function" keyword
Need to assure we test for support of in-line code, for example:

```
function: |  
          function main(params) {
            console.log(params);
            return params || {};
          }
```

we need to have unit tests in place for multiple langs. (e.g., python, java, swift, and php at the very least)

### LINKED ISSUE #745: inline function does not work
My manifest:

```
project:
  name: testfsh
packages:
  p:
    actions:
      echo:
        runtime: nodejs
        function: const main = ({ msg }) => { console.log(msg); return {msg}; }
```

I'm not sure if this is supposed to work or not. I'm getting this error:

```
Error: manifestreader.go [78]: [ERROR_YAML_FILE_FORMAT_ERROR]: File: [manifest.yml]:
==> open /Users/villardl/Projects/whisk/wskp/shell-project-test/const main = ({ msg }) => { console.log(msg); return {msg}; }: no such file or directory
```

pritidesai:
Closes #745 
Closes #622 

This PR includes following changes:

1. Error out if `function` and `code` both exists for an action.
2. Error out if `runtime` is missing when action `code` is specified.
3. Validate `runtime`, if specified runtime is part of the supported runtimes, else if default runtime can be retrieved based on the what is specified.
4. Added unit tests for these checks.
5. Added support of reading action file content over HTTP or HTTPS.
6. When function is set to remote URL, error out of URL is pointing to a remote directory instead of an action file.
7. Updated runtime errors to little more meaningful.
8. Updated manifest_parser unit testing to make sure we add a check on error from ComposeActionsFromAllPackages()
9. Fixed many unit tests in manifest_parser_test.go which were not running before this commit.


pritidesai:
Failing at undeploy, looking into it.

PR #784, opened 2018-03-08
Adding support for package binding from any namespace and declaring public packages

### LINKED ISSUE #248: Add support for sharing a package
[Package sharing documentation](https://github.com/openwhisk/openwhisk/blob/master/docs/packages.md#sharing-a-package)

Could be done by adding the field `shared` under `package`:
```
package:
  name: myPkg
  shared: yes
```

@paulcastro @mrutkows ?

### LINKED ISSUE #621: Support "public" key (flag) on Package
Packaging spec. now acknowledges support for "public" packages (but NOT actions) with the following new key under the Package schema:

Key Name | Required | Value Type | Default | Description
-- | -- | -- | -- | --
public | no | boolean | false | Optional indicator to deploy the package as a “public”   package (requiring no access credentials).

where "public" reflects the "publish" flag on the CLI.

### LINKED ISSUE #762: wskdeploy only supports /whisk.system bindings or github.com packages
In my case I'm dealing with 2 different namespaces, and I'd like to setup a package binding between these 2 namespaces. Is there any reason why only `whisk.system` and `github` are supported right now ?

pritidesai:
Closes #762 
Closes #621 
Closes #248


utils/dependencies.go
```diff
@@ -65,7 +70,7 @@ func NewDependencyRecord(projectPath string,
 }
 
 func LocationIsBinding(location string) bool {
-	if strings.HasPrefix(location, "/whisk.system") || strings.HasPrefix(location, "whisk.system") {
+	if strings.HasPrefix(location, "/"+WHISK_SYSTEM) || strings.HasPrefix(location, WHISK_SYSTEM) || strings.HasPrefix(location, "/") {
```
mrutkows: was not aware we allowed whisk.system without the "/"; is this something allowed in general or just wskdeploy?

pritidesai: doesn't look like `whisk.system` is allowed as is in creating package binding, namespace is always identified with `/`, let me fix this.

```
wsk package bind /whisk.system/samples valhallaSamples --param place Valhalla
ok: created binding valhallaSamples
wsk action invoke --result valhallaSamples/greeting --param name Odin
{
    "payload": "Hello, Odin from Valhalla!"
}
```

PR #789, opened 2018-03-11
Add assets annotation tags when assets exported

kpavel:
Currently exported assets manifest is missing annotations assets have in OpenWhisk.

### DIFF
--- a/parsers/yamlparser.go
+++ b/parsers/yamlparser.go
@@ -19,6 +19,7 @@ package parsers
 
 import (
 	"github.com/apache/incubator-openwhisk-client-go/whisk"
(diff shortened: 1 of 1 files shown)


mrutkows:
@kpavel @pritidesai why would we filter out managed annotations?  If they are removed on export, then their usefulness is lost to external tooling (e.g., IDEs) which the managed feature was authored for.

kpavel:
@mrutkows  the idea is to export all the annotations asset already has, excluding the "MANAGED" annotation. We should omit it to prevent future deployments ambiguity, e.g. user deploys same manifest without --managed on another OW instance. If user deploys it with --manage, it will use managing project name which we already export to the manifest. The MANAGE annotation will be added automatically to the assets in the new deployment.

PR #801, opened 2018-03-14
Adding Docker and Native support

### LINKED ISSUE #279: Support new "native" (or "binary") kind
See PR for this feature...
https://github.com/apache/incubator-openwhisk/pull/2112

Need to support in spec. and in tooling.

### LINKED ISSUE #623: Support "blackbox" Docker images
This has been a discussion since last Sept. and asked Daisy to look into supporting at that time.  Need to track this work via this issue.

```
Sept 2016:
Daisy: Blackbox (Docker) Actions: Need to think about how to set the field "runtime" if it is a docker action. We should document it.
```

pritidesai:
Closes #623
Closes #279 

List of changes in this PR:

1. Major Refactoring of `manifest_parser.go`:
* Replace `str != ""` kind of check with `len(str) != 0`
* Replace "/" with constant `PATH_SEPARATOR`
* Created a common function to parse inputs and outputs, updated all `Compose*` routines to utilize this common function. Same with annotations.
* Split `ComposeAction` function into smaller routines with validating code, validating function, reading code, reading function, creating action exec depending on action function and code, etc.
2. Added support for docker and native actions with unit test and integration test.


parsers/manifest_parser_test.go
```diff
@@ -893,6 +896,42 @@ func TestComposeActionsForFunctionWithRemoteDir(t *testing.T) {
 	assert.NotNil(t, err, "Compose actions should have exited with error when code is specified but runtime is missing.")
 }
 
+// validate manifest_parser.ComposeActions() method
+func TestComposeActionsForDocker(t *testing.T) {
```
mrutkows: Nice unit test we have needed for some time...

parsers/manifest_parser_test.go
```diff
@@ -893,6 +896,42 @@ func TestComposeActionsForFunctionWithRemoteDir(t *testing.T) {
...
+	var expectedResult, actualResult string
+	for _, action := range actions {
+		switch action.Action.Name {
+		case "OpenWhiskSkeleton":
+		case "OpenWhiskSkeletonWithNative":
+			assert.Equal(t, utils.BLACKBOX, action.Action.Exec.Kind, fmt.Sprintf(TEST_MSG_ACTION_DOCKER_KIND_MISMATCH, action.Action.Exec.Kind))
+			assert.Equal(t, NATIVE_DOCKER_IMAGE, action.Action.Exec.Image, TEST_MSG_ACTION_DOCKER_IMAGE_MISMATCH)
+		case "CustomDockerAction1":
+		case "CustomDockerAction2":
+			expectedResult, _ = filepath.Abs("../tests/src/integration/docker/actions/exec.zip")
```
mrutkows: Would have preferred this declared at top of function so we can quickly identify test data files used within the test method and they do not scroll off the right edge.

parsers/yamlparser.go
```diff
@@ -90,24 +90,26 @@ type YAMLParser struct {
...
+	ExposedUrl  string                 `yaml:"exposedUrl"`
+	Webexport   string                 `yaml:"web-export"`
+	Main        string                 `yaml:"main"`
+	Docker      string                 `yaml:"docker,omitempty"`
+	Native      bool                   `yaml:"native,omitempty"`
+	Limits      *Limits                `yaml:"limits"`
+	Inputs      map[string]Parameter   `yaml:"inputs"`
+	Outputs     map[string]Parameter   `yaml:"outputs"`
 	Annotations map[string]interface{} `yaml:"annotations,omitempty"`
 	// TODO() this is propoagated from package to every action within that package
```
mrutkows: we will get you yet typo! ...

tests/src/integration/docker/actions/go/sample.go
```diff
@@ -0,0 +1,44 @@
+/*
```
mrutkows: Do we need the source code for the "exec" binary?

tests/dat/manifest_data_compose_actions_for_docker.yaml
```diff
@@ -0,0 +1,39 @@
...
+              native: true
+          CustomDockerAction3:
+              function: ../src/integration/docker/actions/bash
+              native: true
+          CustomDockerAction4:
+              function: ../src/integration/docker/actions/go
+              native: true
+          CustomDockerAction5:
+              function: ../src/integration/docker/actions/go
+              docker: mydockerhub/myimage
```
ddragosd: great 👍 

should we assume that custom/private docker registries are supported as well, provided the machine has the credentials needed to connect to that registry ?

tests/src/integration/docker/actions/go/sample.go
pritidesai: This is a sample in Go from https://console.bluemix.net/docs/openwhisk/openwhisk_actions.html#creating-go-actions. I have three examples included, (1) a dir with exec in bash (function: <dir>) (2) a dir with exec in go (compiled) (function: <dir>) (3) exec.zip. Its pretty much same workflow for example.c

tests/dat/manifest_data_compose_actions_for_docker.yaml
pritidesai: @ddragosd yes and no, custom docker registries are supported but not private. OpenWhisk does not support private registries yet (see https://github.com/apache/incubator-openwhisk/issues/1362) since the docker image is pulled when you invoke that action for the first time so even if the machine from where the action was created has credentials, its not useful.

parsers/manifest_parser_test.go
pritidesai: 👍 done

mrutkows:
@ddragosd @pritidesai In order to support private (with credentials), we would need to add support for thee credential and repository schema (envisionsed in spec.).  Priti, do you agree with this assessment?  If so, can we open an epic/issues for this and prioritize?

csantanapr:
This is not supported in core, the invoker doesn’t deal with credentials to pull the image it just executes docker pull 

Cc @rabbah

pritidesai:
@mrutkows I think its beyond the scope of `wskdeploy`. We have repository and credentials defined [here](https://github.com/apache/incubator-openwhisk-wskdeploy/blob/master/specification/openwhisk_v0.9.2.pdf) on line 558 and line 571. But with docker actions the way they are designed, images are not pulled during action creation. So even if we have access to credentials of private repo and we could pass that to OpenWhisk server but the its not supported by server side. 

We can address this by what Rodric has commented here https://github.com/apache/incubator-openwhisk/issues/1362 saying "_One way to support this is by pulling images from the private registry to a whisk managed registry at action create time. This would avoid some the challenges around key management for docker login. In general I think this is a good approach for docker actions as it will make it easier for vetting images; but not without its demerits._" We can definitely add support for this in `wskdeploy`, is that what we are looking for?

csantanapr:
Yep this is something that wskdeploy can’t solve at the moment

PR #802, opened 2018-03-14
WIP: Resolves #730

### LINKED ISSUE #730: Multiple projects management
Currently it is impossible to use wskdeploy to  deploy OW assets (action, package...etc) of the same signature from different manifests having different project names. The problem is that wskdeploy overrides the "managed project" annotation. The workaround could be to use different package names, which results in assets duplication. Duplication is undesirable, because it complicates maintenance. 
Introducing a feature allowing to manage OW assets by multiple projects would help to avoid it.
Suggestion:
- Use different annotation key instead of "manage" for the sake of backward compatibility. E.g. `managedList` holding an array of managed project names.
- On manifest deployment, if a deployed asset already exists in OW, check whether the manifest asset is identical to the existing one.
If not: override the existing asset with the manifest one (similar to the current functionality)
If yes: add the project specified in the manifest to the `managedList`.

kpavel:
This PR extends wskdeploy with a capability to explicitly define relationships between projects and assets beyond single project. More specifically, it allows to express one (asset) to many (projects) relationships, which are useful in e.g., reusing assets managed by one project as “library assets” in multiple other projects. The design goals in this PR are: (a) to preserve current functionality of wskdeploy, (b) not to change current manifest structure, (c) avoid assets duplication among projects while having a single project, the project where an asset have been originally defined as managed, as a single point of truth for any reference to this asset made from other projects.
  

Relationships of projects and assets are expressed via a new MANAGED_LIST annotation of an asset. This  functionality is consumed by the user via a (-r) wskdeploy attribute. It receives a YAML file with a format similar to ordinary wskdeploy manifest.yaml or deployment.yaml manifest files.

This PR gains:
- an option to export only a subset of existing managed assets by specifying them in a relationships manifest file of another project, which will result in this project name to be added to the MANAGED_LIST annotation. The new project name can be used in calling wskdeploy for exporting assets of that new project. 
- assets reuse by different projects with central asset management by its own single MANAGER project
- avoids unnecessary assets duplication

Usage example:
- 1) Deploy project A with a  set of managed assets {a}
- 2) Deploy project B with a different set of managed assets {b}
- After steps  1 and 2, there will be a union of  assets {a} and{b} in OW with MANAGED annotation pointing to the project managing the asset

- 3) Deploy project C with the manifest containing nothing but the project name + a relationships manifest file pointing to assets from both {a} and {b} sets.
- Step 3  results in no new assets creation, but rather adds a new entry to the  MANAGED_LIST annotations array. In this case, C will be added as a project name to the MANAGED_LIST of the assets meaning that C uses these assets. - 4) Export project C. Notice that the  manifest file created by the export operation comprisesboth assets from {a} and {b} without any reference to  to the A and B projects.

- 5) Undeploy project C. This results in no assets deletion, but rather removing of the relationship to project C from the assets MANAGED_LIST annotation.
- 6) Undeployment of project A or B will result in complete removal of the {a} or {b} assets sets, although in the future we may  consider adding an explicit --force flag to avoid undesired undeployments.


pritidesai:
@kpavel please create a separate branch named `fix-for-730` instead of proposing changes in your `master`, reviewing off of master branch is not feasible. Also, will you be able to please explain the use case and your approach to it in our next week's Spec meeting?

kpavel:
@pritidesai Opened new PR #804  with different head branch

PR #804, opened 2018-03-15
Fix for 730

kpavel:
This PR extends wskdeploy with a capability to explicitly define relationships between projects and assets beyond single project. More specifically, it allows to express one (asset) to many (projects) relationships, which are useful in e.g., reusing assets managed by one project as “library assets” in multiple other projects. The design goals in this PR are: (a) to preserve current functionality of wskdeploy, (b) not to change current manifest structure, (c) avoid assets duplication among projects while having a single project, the project where an asset have been originally defined as managed, as a single point of truth for any reference to this asset made from other projects.
  

Relationships of projects and assets are expressed via a new MANAGED_LIST annotation of an asset. This  functionality is consumed by the user via a (-r) wskdeploy attribute. It receives a YAML file with a format similar to ordinary wskdeploy manifest.yaml or deployment.yaml manifest files.

This PR gains:
- an option to export only a subset of existing managed assets by specifying them in a relationships manifest file of another project, which will result in this project name to be added to the MANAGED_LIST annotation. The new project name can be used in calling wskdeploy for exporting assets of that new project. 
- assets reuse by different projects with central asset management by its own single MANAGER project
- avoids unnecessary assets duplication

Usage example:
- 1) Deploy project A with a  set of managed assets {a}
- 2) Deploy project B with a different set of managed assets {b}
- After steps  1 and 2, there will be a union of  assets {a} and{b} in OW with MANAGED annotation pointing to the project managing the asset

- 3) Deploy project C with the manifest containing nothing but the project name + a relationships manifest file pointing to assets from both {a} and {b} sets.
- Step 3  results in no new assets creation, but rather adds a new entry to the  MANAGED_LIST annotations array. In this case, C will be added as a project name to the MANAGED_LIST of the assets meaning that C uses these assets. - 4) Export project C. Notice that the  manifest file created by the export operation comprisesboth assets from {a} and {b} without any reference to  to the A and B projects.

- 5) Undeploy project C. This results in no assets deletion, but rather removing of the relationship to project C from the assets MANAGED_LIST annotation.
- 6) Undeployment of project A or B will result in complete removal of the {a} or {b} assets sets, although in the future we may  consider adding an explicit --force flag to avoid undesired undeployments.


pritidesai:
@kpavel I am still trying to understand the use case and your approach, some thoughts while I am learning:

Why introducing a new `-r` attribute? Can we define this whole dependency map and library concept under `Dependency` section? Like @lionelvillard suggested in one of the comments in the issue #730, we can add MANAGED annotation in dependency as well if its not already supported. Or we create a separate sub section under `dependency` specifying project name and if the entities referred in extension manifest yaml are not defined in the same yaml than check the dependency section for any library kind of dependency chain.

library manifest.yaml:

```
project:
    name: LIB
    packages:
        lib_package:
            version: 1.0
            license: Apache-2.0
            actions:
                lib_greeting_1:
                    version: 1.0
                    function: src/greeting.js
                    runtime: nodejs:6
                lib_greeting_2:
                    version: 1.0
                    function: src/greeting.js
                    runtime: nodejs:6
            triggers:
                lib_trigger:
            rules:
                lib_rule:
                    trigger: lib_trigger
                    action: lib_greeting_1
```

extension manifest.yaml:

```
project:
    name: EXT
    packages:
        triggerrule_ext:
            version: 1.0
            license: Apache-2.0
            dependency:
                library: LIB
            actions:
                ext_greeting:
                version: 1.0
                function: src/greeting.js
                runtime: nodejs:6
            rules:
                ext_rule:
                    trigger: lib_trigger
                    action: ext_greeting
```

thoughts?

davidbreitgand:
@pritidesai My two cents: "dependency" will become  an over-loaded term if we use it beyond describing dependencies on packages from github and whisk.system. For that reason we thought of using "relationships" and defining all relationships between projects explicitly at the granularity of an asset and not necessarily a package. IMHO, introducing a separate [optional] relationship manifest minimizes impact on the existing code and gets us backward compatibility, while the alternative (although definitely doable and reasonable) has much more impact on the current code base.

mrutkows:
@davidbreitgand I do not like introducing a -r flag and somehow differentiating from static or dynamically linked dependent packages.  IMO, and after discussing with Priti, we should track all depenedencies using the managed annotations.  There should be no confusion to the end user as dependencies that are referenced at a location are given a local name and bound and there is not external location they are searched for in the public packages and bound to a local name as well.  This should solve any use case.

kpavel:
Following discussion reopened new PR #813 providing new implementation

PR #813, opened 2018-03-22
Provides support to export projects with dependencies

### LINKED ISSUE #730: Multiple projects management

kpavel:
Resolves #730
- Provides support to export projects with dependencies
- Optimizes rule export (leveraging added rule annotations)
- Adds MANAGED annotation to dependencies in OW package bindings

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -44,18 +44,7 @@ var exportCmd = &cobra.Command{
 
 var config *whisk.Config
 
-func ExportRule(wskRule whisk.Rule, pkgName string, maniyaml *parsers.YAML) {
-	if maniyaml.Packages[pkgName].Rules == nil {
-		pkg := maniyaml.Packages[pkgName]
-		pkg.Rules = make(map[string]parsers.Rule)
-		maniyaml.Packages[pkgName] = pkg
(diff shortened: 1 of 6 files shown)


pritidesai:
Please do not merge this.

mrutkows:
Closing in favor of agreed-upon impl. as merged in: https://github.com/apache/incubator-openwhisk-wskdeploy/pull/816

PR #817, opened 2018-03-23
add notification webhook for slack notification of PR results.

dgrove-oss:


dgrove-oss:
Verified as working.
screen shot 2018-03-27 at 11 03 26 am

PR #827, opened 2018-03-28
Updated documentation and examples to use the new packages syntax

### LINKED ISSUE #826: Fix documentation that uses the old package notation instead of packages
The current programming guide uses the old `package` notation rather than the new `packages` one. It should be updated to reflect changes and ensure that readers can run the examples given.

brunogirin:
Should fix #826 

Updated the syntax from:

```yaml
package:
  name: package_name
  version 1.0
...
```

to:

```yaml
packages:
  package_name:
    version: 1.0
...
```
in all docs and example files. I'm not sure about the file that starts `project:` so this may need adjusting too.

### DIFF
--- a/docs/examples/deployment_hello_world_triggerrule_unbound.yaml
+++ b/docs/examples/deployment_hello_world_triggerrule_unbound.yaml
@@ -15,7 +15,7 @@
 #
 
 project:
-  package:
+  packages:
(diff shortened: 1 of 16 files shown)


mrutkows:
@brunogirin Thanks Bruno!

PR #830, opened 2018-03-28
Fix the build failure in building binaries

houshengbo:
This PR updates the dependency in gogradle, so that it will build based on the latest library.
We are so far consolidating the dependencies in Godeps and gogradle, so they still co-exist at this time.


pritidesai:
Yup indeed, Thanks Vincent for addressing this, will add a disclaimer in go deps and go Gradle that they both have to be updated, is there any solution to enforce this doesn’t happen in future?

houshengbo:
We are deprecating Godeps, which means we will build the binaries purely on gogradle, and remove the makefile.

PR #844, opened 2018-04-03
Add the LICENSE and NOTICE into the binary packages

houshengbo:


daisy-ycguo:
I will suggest this PR to be merged after my PR #842. Or it will cause conflict.

pritidesai:
`wski18n/i18n_resources.go` is auto generated file, can we opt out of the mandatory header requirement for this file if possible? otherwise, we might want to add a note somewhere explicitly specifying to add header after resources file is generated.

houshengbo:
@pritidesai For i18n_resources.go, wskdeploy already generates it before we release anything, so users do not have to generate it. As long as we have it in our code base, we need to add the header.

mrutkows:
@houshengbo surely, there is an Apache exception for generated files.... after all they are not "licensed" by an individual...

mrutkows:
See tracking issue for the i18n_resources.go issue (i.e., where to document as an exclusion):
https://github.com/apache/incubator-openwhisk-release/issues/94

PR #856, opened 2018-04-05
Add the tag into the names of the binary packages

houshengbo:


.travis.yml
```diff
@@ -62,6 +62,12 @@ before_deploy:
 - echo "Deploying $RELEASE_PKG_FILE to GitHub releases."
 - export GIT_TAG="latest"
 - export TAG=false
+- if [ ! -z "$TRAVIS_TAG" ] ; then
```
mrutkows: : If the current build is for a git tag, this variable is set to the tag’s name.

.travis.yml
```diff
@@ -62,6 +62,12 @@ before_deploy:
...
+      export GIT_TAG=$TRAVIS_TAG;
+      export TAG=true;
+  fi
+- ./gradlew --console=plain releaseBinaries -PpackageVersion=$GIT_TAG
```
mrutkows: So we now invoke the existing gradle task...

mrutkows:
@pritidesai please review as I recall the last time we invoked this gradle task there was an issue perhaps that you encountered.

PR #864, opened 2018-04-10
upgrading go client version

### LINKED ISSUE #810: wskdeploy assumed "guest" namespace if one is not set
if you have a .wskprops file without a NAMESPACE defined, it appears wskdeploy assumes "guest" as the namespace which is incorrect. It should use "_" or do an actual look up and retrieve the namespace.

```
> wskdeploy export
Info: The API host is [https://openwhisk.ng.bluemix.net], from .wskprops.
Info: The auth key is set, from .wskprops.
Info: The namespace is [guest], from .wskprops.
Info: Unmarshal OpenWhisk runtimes from local values.
Error: The supplied authentication is not authorized to access 'guest'. (code 2898491)
```

pritidesai:
Setting default namespace to `_` instead of guest. Changes were made in go client, upgrading dependency file here.

Closes #810


pritidesai:
@beemarie once this PR is merged, I will create a separate release `0.9.3.1` so that you can integrate the latest changes in your template work.

beemarie:
@pritidesai This fix is still a workaround, right?  If the namespace in the manifest.yaml were set to something other than '_', it wouldn't get read in from the manifest, and would be set to the default.

build.gradle
```diff
@@ -62,8 +62,8 @@ dependencies {
         build(['name':'github.com/nicksnyder/go-i18n/i18n/bundle', 'version':'991e81cc94f6c54209edb3192cb98e3995ad71c1', 'transitive':false])
         build(['name':'github.com/nicksnyder/go-i18n/i18n/language', 'version':'991e81cc94f6c54209edb3192cb98e3995ad71c1', 'transitive':false])
         build(['name':'github.com/nicksnyder/go-i18n/i18n/translation', 'version':'991e81cc94f6c54209edb3192cb98e3995ad71c1', 'transitive':false])
-        build(['name':'github.com/apache/incubator-openwhisk-client-go/whisk', 'version':'1759868a61729708a8ab1a6de1b409d6e2aea00a', 'transitive':false])
-        build(['name':'github.com/apache/incubator-openwhisk-client-go/wski18n', 'version':'1759868a61729708a8ab1a6de1b409d6e2aea00a', 'transitive':false])
+        build(['name':'github.com/apache/incubator-openwhisk-client-go/whisk', 'version':'defaf9776014c3d9edbffb2671ed166855d0ee60', 'transitive':false])
```
mrutkows: assume we need some new go client function and forced a drag on i18n.  However, our functional code did not change to add any new features?

mrutkows: ahh, so this is behavior we get from Go client...

PR #887, opened 2018-05-02
WIP: please do not merge: an attempt to cleanup list of deps

pritidesai:
Updating list of dependencies to remove ones which are not used in whisk deploy source code.


pritidesai:
I am reopening it as since I have started experimenting with this, lets run through and validate all the deps we have are needed or not, I am adding deps back for which travis build failed.

pritidesai:
need to revisit this with may be better tooling to catch cross references ...

PR #894, opened 2018-05-03
adding list of YAML key/values which supports env. variables

### LINKED ISSUE #892: Create markdown page describing where Interpolation is supported
This issue #[891](https://github.com/apache/incubator-openwhisk-wskdeploy/issues/891) 
reminded me we discussed documenting all the places in the YAML we interpolate.  

Hopefully, the list can easily be derived from grepping for where we use the function "_InterpolateStringWithEnvVar_" and documenting the cases (where in schema) we provide Interp.

My comments were reflected in the issue above as:
```
... we have been adding interpolation for keys/values in many places (and thankfully the way the code/functions are structure has been relatively simple); however, we should likely create/publish that lists all the places we support Interpolation (from env. vars) and link to it from this page here: https://github.com/apache/incubator-openwhisk-wskdeploy/blob/master/docs/wskdeploy_action_env_var_parms.md#actions.
```

pritidesai:
Closes #892

### DIFF
--- a/docs/wskdeploy_interpolation.md
+++ b/docs/wskdeploy_interpolation.md
@@ -1,0 +1,130 @@
+<!--
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for additional
+# information regarding copyright ownership.  The ASF licenses this file to you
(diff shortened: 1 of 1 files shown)


pritidesai:
👍

PR #914, opened 2018-05-18
Fix mini license header

pritidesai:


pritidesai:
@mrutkows I am extremely sorry for proposing invalid headers in all files under wskdeploy tests folder. Please verify these changes with scan code and let me know if there is anything i need to fix.

PR #943, opened 2018-05-29
Add the help message, if manifest file is not accesible to wskdeploy and undeploy

houshengbo:


cmd/root.go
```diff
@@ -146,6 +146,9 @@ func loadDefaultManifestFileFromProjectPath(command string, projectPath string)
 	} else {
 		stderr = wski18n.T(wski18n.ID_ERR_MANIFEST_FILE_NOT_FOUND_X_path_X,
 			map[string]interface{}{wski18n.KEY_PATH: projectPath})
+		if cmd != nil {
+			cmd.Help()
```
pritidesai: With these changes, the help message is displayed before the actual error which seems out of order.

image

Can we change this to the following way of combining usage and the error:

```
if cmd != nil {
			stderr = stderr + wski18n.T(cmd.UsageString())
		}
```

This is how it appears:

image

But with this, everything shows in red as an error, please change it so that its displayed as a regular text.

rabbah:
Thanks @houshengbo - it would be good practice to add go unit tests for these improvements as they’re developed.

pritidesai:
Created separate low priority issue for the same, #947

PR #946, opened 2018-05-31
runtimes

### LINKED ISSUE #927: action deployed when runtime is deprecated
I can reproduce by deploying openwhisk with the following manifest:

```
> cat ansible/files/runtimes.json
{
    "runtimes": {
        "nodejs": [
            {
                "kind": "nodejs",
                "image": {
                    "name": "nodejs6action"
                },
                "deprecated": true
            },
            {
                "kind": "nodejs:6",
                "default": true,
                "image": {
                    "name": "nodejs6action"
                },
                "deprecated": false
            }
        ]
    },
    "blackboxes": [
        {
            "name": "dockerskeleton"
        }
    ]
}
```

and the following wskdeploy manifest
```
packages:
  p:
    actions:
      hello:
        function: hello.js
        runtime: nodejs
```

this is the output from `wskdeploy`:

```
Info: The API host is [https://localhost], from .wskprops.
Info: The auth key is set, from .wskprops.
Info: The namespace is [_], from .wskprops.
Info: Unmarshal OpenWhisk runtimes from local values.
Warning: The mandatory key [package version] should be set. Using default value [0.0.1]...
Warning: The value for key [package version] is not saved in the current version of wskdeploy.
Warning: The mandatory key [package license] should be set. Using default value [unlicensed]...
Warning: The value for key [package version] is not saved in the current version of wskdeploy.
Warning: Invalid or missing runtime [nodejs] specified in manifest for the action [hello].
Warning: Runtime changed to [nodejs:6] based on the action's source file extension for action [hello].
Info: Deploying package [p] ...
Info: package [p] has been successfully deployed.

Info: Deploying action [p/hello] ...
Info: action [p/hello] has been successfully deployed.
Success: Deployment completed successfully.
```

Note the `Warning: Invalid or missing runtime [nodejs] specified in manifest for the action [hello].` is correct but I think the action should not automatically deploy without an override from the user as there may be incompatible or breaking changes relative to the source code.

*is there a way to disable this automatic remapping on the command line?*

### LINKED ISSUE #928: hardcoded protocol does not allow for http to be used
this line hardcodes `https` as the protocol.

https://github.com/apache/incubator-openwhisk-wskdeploy/blob/b9fd4d317888319ddd3fea3ad8458b2b661ec407/utils/runtimes.go#L85

if the api host already includes a protocol, wskdeploy fetches the runtime manifest from hardcoded values here:
https://github.com/apache/incubator-openwhisk-wskdeploy/blob/b9fd4d317888319ddd3fea3ad8458b2b661ec407/utils/runtimes.go#L229

this is startling - instead `wskdeploy` should fail --- the hardcoded values could be very different from the deployment wrt action default, supported actions, etc... isn't this dangerous?!

### LINKED ISSUE #929: if manifest read fails, wskdeploy should not read details from hardcoded manifest
wskdeploy fetches the runtime manifest from hardcoded values here should it fail reading from the API host.

https://github.com/apache/incubator-openwhisk-wskdeploy/blob/b9fd4d317888319ddd3fea3ad8458b2b661ec407/utils/runtimes.go#L229

this is startling - instead `wskdeploy` should fail --- the hardcoded values could be very different from the deployment wrt action default, supported actions, etc... isn't this dangerous?!

### LINKED ISSUE #930: wskdeploy should fail if host info does not parse or cannot be retrieved
This method should exist `wskdeploy` if there is an error - it should not proceed and surely not silently without even warning you of the error.

https://github.com/apache/incubator-openwhisk-wskdeploy/blob/c3459f0c200eaea8e63b5bf7b6e75e852adb067d/cmd/root.go#L123-L128

which can happen if this line fails
https://github.com/apache/incubator-openwhisk-wskdeploy/blob/b9fd4d317888319ddd3fea3ad8458b2b661ec407/utils/runtimes.go#L124

### LINKED ISSUE #931: schema encoding of the runtime manifest schema uses uint16
https://github.com/apache/incubator-openwhisk-wskdeploy/blob/b9fd4d317888319ddd3fea3ad8458b2b661ec407/utils/runtimes.go#L48-L52

why is this uint16 limit imposed; is there a concern about size of the struct? i suggest unit... the manifest schema doesn't specify a limit.

fwiw as I've noted for the CLI and go SDK - the encoding of these types as structs in mostly unnecessary.

pritidesai:
Closes #931 
Closes #930 
Closes #929 
Closes #928 
Closes #927


rabbah:
Thanks @pritidesai for the all these patches and enhancements. I will review and try with my local dev environment.

PR #950, opened 2018-06-04
Continue spec. conversion to markdown and add navigation

mrutkows:


specification/html/spec_actions.md
```diff
@@ -19,87 +19,44 @@
...
+| final | no | boolean | false | The optional flag (annotation) which makes all of the action parameters that are already defined immutable.<p><b>Note</b>: this option is ONLY valid if <em>"web"</em> or <em>"web-export"</em> is set to <em>‘true’</em>.<p> |
+| web-custom-options | no | boolean | false | The optional flag (annotation) enables a web action to respond to OPTIONS requests with customized headers, otherwise a [default CORS response](https://github.com/apache/openwhisk/blob/master/docs/webactions.md#options-requests) applies. |
+| require-whisk-auth | no | boolean | false | The optional flag (annotation) protects the web action so that it is only accessible to an authenticated subject. |
+| main | no | string | N/A | The optional name of the function to be aliased as a function named “main”.<p><em><b>Note</b>: by convention, Action functions are required to be called “main”; this field allows existing functions not named “main” to be aliased and accessed as if they were named “main”.</em></p>|
+
+### Deprecated Fields
+
+The following fields are considered deprecated as of the current specification version.
```
pritidesai: `Code` is not deprecated, its one of the supported fields and required if `function` is not specified. When a `code` is specified, `runtime` becomes mandatory field.

mrutkows: @pritidesai corrected and also added new req. (and to .pptx that I will need to upload at the end of this conversion so they reflect each other).

Issue #975, opened 2018-08-31
HTTP response documentation


brunogirin:
The current documentation doesn't include any example that make use of HTTP responses. It would be good to add one that shows:
- what to change in the API definition
- what to change in the action itself and in particular:
    - how to return a success or an error structure using plain JSON
    - how to do the same using `Promise.resolve` and `Promise.reject`

brunogirin:
@pritidesai can I close this issue now that you merged the PR?

pritidesai:
thanks @brunogirin yes we can close this.

PR #976, opened 2018-08-31
#975 Added HTTP response documentation

brunogirin:
Added a new documentation page that shows how to create and deploy actions that use the HTTP response, including:
- changes to the manifest,
- changes to the action itself showing a raw JSON solution and a Promise based solution


brunogirin:
Apologies for the multiple updates to that PR, there were a few issues with the initial code and I then decided to add one more page to include a sequence + HTTP response example. It should now be good for review.

PR #983, opened 2018-09-11
adding ruby runtime

pritidesai:
Closes #982 
Closes #970

PR #985, opened 2018-09-12
Validate that there no perfix before adding https

kpavel:


pritidesai:
Closing, since its duplicate of PR #983

PR #996, opened 2018-10-18
Enable export verbose output

kpavel:
When using "project export", similar to project deployment, verbose stdout should be supported


pritidesai:
@kpavel the build failing with syntax error, there are additional spaces in front of two lines you have added.

PR #997, opened 2018-10-24
Upgrade the Go version to 1.9

houshengbo:


.travis.yml
```diff
@@ -4,7 +4,7 @@
 sudo: required
 language: go
 go:
-- 1.8
+- 1.9
```
csantanapr: what about using `1.11` ?

houshengbo: 1.11 will fail at this time. Still need to check what needs to be changed for higher version.

PR #1008, opened 2018-11-06
Upload PDF of 0.9.3 spec.

mrutkows:


csantanapr:
We should not have the spec in a proprietary form like PDF, is impossible to see a git diff of the changes

Spec should be a file in text, markdown or html or plain text, or combination

pritidesai:
yup agree @csantanapr, we are in process of converting it into markdown https://github.com/apache/incubator-openwhisk-wskdeploy/tree/master/specification#package-specification but dont want to loose any changes until we have completed the conversion.

csantanapr:
How can I track progress on the conversion is there an Issue or PR I can subscribe ?

Issue #1014, opened 2018-11-28
wskdeploy failing to read configuration file from user home directory for 386 version


jthomas:
The 386 release of the tool fails to retrieve the user's home directory from the system. This causes it to expect the `wskprops` file in the current directory rather than the user's home directory. 

## example 

- Run a `docker run -it ubuntu` container and retrieve the latest release archive of the tool from the release page.

- Using a simple manifest run the 386 & amd64 versions of the tool with the `-v` flag and check the different `CfgFile` property values.

```
root@8651a121af02:~/wskdeploy/linux/amd64# ./wskdeploy -v -m manifest.yaml
Info: deploy using deployment file []...

Info: Configuration:
      > ApiHost: []
      > Auth: []
      > Namespace: []
      > ApiVersion: []
      > CfgFile: [/root/.wskprops]
      > CliVersion: [latest]
      > ProjectPath: [.]
      > DeploymentPath: []
      > ManifestPath: [manifest.yaml]
      > Preview: [false]
      > Strict: [false]
      > Key: []
      > Cert: []
      > Managed: [false]
      > ProjectName: []
      > ApigwAccessToken: []
      > Verbose: [true]
      > Trace: [false]
      > Sync: [false]
      > Report: [false]
      > Param: [[]]
      > ParamFile: []
```

```
 ./wskdeploy -v -m ../amd64/manifest.yaml
Info: deploy using deployment file []...

Info: Configuration:
      > ApiHost: []
      > Auth: []
      > Namespace: []
      > ApiVersion: []
      > CfgFile: [.wskprops]
      > CliVersion: [latest]
      > ProjectPath: [.]
      > DeploymentPath: []
      > ManifestPath: [../amd64/manifest.yaml]
      > Preview: [false]
      > Strict: [false]
      > Key: []
      > Cert: []
      > Managed: [false]
      > ProjectName: []
      > ApigwAccessToken: []
      > Verbose: [true]
      > Trace: [false]
      > Sync: [false]
      > Report: [false]
      > Param: [[]]
      > ParamFile: []
```

## cause

The cause is due to `user.Current` method [not being available](https://github.com/apache/incubator-openwhisk-wskdeploy/blob/master/utils/misc.go#L65-L67) for 386.

Here's a test file to show the issue...

```go
package main

import "fmt"
import "os"
import "os/user"

func main() {
	usr, err := user.Current()
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
[...]

PR #1015, opened 2018-11-28
Quickfix: Add fallback method to find wskprops when go-client fails

mrutkows:
We need a longer-term fix that starts in the Go Client (which we use in some cases) where Go client needs to be made more robust and perhaps look to packages such  [ (i.e., https://github.com/mitchellh/go-homedir) it appears to be a single file under MIT license which if we switch to we would need to note.](https://github.com/mitchellh/go-homedir).

Which is used by the CLI and appears to be a single file under MIT license which if we switch to we would need to note.  This means go-client, and CLI may need to be cleaned up and once that is done wskdeploy would strictly rely upon go-client ReadProp() impl.

### DIFF
--- a/utils/misc.go
+++ b/utils/misc.go
@@ -60,10 +60,19 @@ type RuleRecord struct {
 	Packagename string
 }
 
+func fallbackHome() string {
+	if home := os.Getenv("HOME"); home != "" {
+		return home
+	}
+	// For Windows.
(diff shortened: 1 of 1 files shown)


mrutkows:
quick fix to resolve: https://github.com/apache/incubator-openwhisk-wskdeploy/issues/1014
Fixes #1014

PR #1017, opened 2018-11-29
Deleting static runtimes data file, instead, adding data to go file

### LINKED ISSUE #1005: Action has no kind set
When attempting to execute `wskdeploy -m manifest.yml`, I get an error: "Action [action_name] has no kind set.

image

See manifest.yaml below

```yaml
packages:
    helloworld:
        actions:
            helloJS:
                function: src/hello.js
                runtime: nodejs:8
            hello-world-nodejs-delay:
                function: src/hello-delay.js
                runtime: nodejs:8

```

I have tried every version of **wskdeploy** back to v09.7

pritidesai:
There is no way in Golang to bundle data file at compile time and read from it during the runtime except using `packr` to package/bundle files in the binary itself. Simplifying the implementation by moving data from a separate json file to go file where the data is being read.

Closes #1005


pritidesai:
We are merging this PR but please address this issue https://github.com/apache/incubator-openwhisk-wskdeploy/issues/1018 as a high priority.

runtimes/runtimes.go
```diff
@@ -56,7 +54,6 @@ const (
 	HTTP_CONTENT_TYPE_VALUE = "application/json; charset=UTF-8"
 	RUNTIME_NOT_SPECIFIED   = "NOT SPECIFIED"
 	BLACKBOX                = "blackbox"
-	RUNTIMES_FILE_NAME      = "runtimes.json"
```
mrutkows: I hate getting rid of this...

PR #1019, opened 2018-12-13
Add support for .Net dotnet:2.2 action kind

shawnallen85:
Updated items to meet requirements of https://github.com/apache/incubator-openwhisk/issues/4177


tests/src/integration/runtimetests/manifest.yaml
```diff
@@ -158,6 +162,9 @@ packages:
             helloworldjava-without-explicit-runtime:
                 function: src/hello.jar
                 main: Hello
+            helloworlddotnet22-without-explicit-runtime:
```
csantanapr: I think we don't need this entry, you can remove it.
For dotnet is always explicit because is only .zip, there is no way to deduce the runtime using .zip

PR #1023, opened 2019-01-09
Enable programatic support for additional request headers

mdeuser:


mdeuser:
@pritidesai thanks for taking a look.  after hunting around, i think i found a place where the unit test could be added.  if it looks good, it's ready for merging.  thanks!

PR #1029, opened 2019-02-04
Add API entity schema

mdeuser:


specification/html/spec_apis.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Grammar
+
+```yaml
+<apiName>:
+    <basePath>:
+        <relativePath>:
+            <actionName>:
+                method: <get | delete | put | post | ...>
+                response-type: <http | json | text | html>
```
pritidesai: is this `response-type` or `response`? similarly few more places in the rest of the doc.

specification/html/spec_apis.md
```diff
@@ -0,0 +1,122 @@
...
+              response-type: json
+```
+
+### Requirements
+
+- The API name MUST be less than or equal to 256 characters.
+- The API `basePath` value MUST start with a `/` character.
+- The APi `relativePath` value MUST start with a `/` character.
+- The API entity schema includes all required fields declared above.
+- Only web actions, actions having `web-export` set to `true`, can be used as an API endpoint's action.
```
pritidesai: please add a note saying actions are converted to web actions if non web actions are used in the api section

specification/html/spec_apis.md
mdeuser: it should be `response`.  i'll change all of these. :+1:

Issue #1033, opened 2019-02-18
memorySize is limited to 512MB


TPei:
wskdeploy limits memory size to 512mb. To me, it is unclear why that is. I'd propose raising those limits, OpenWhisk (at least on IBM) supports up to 2048MB limits

TPei:
Hmm, ok, according to the documentation, the memory limit is still 512MB https://github.com/apache/incubator-openwhisk/blob/master/docs/reference.md

TPei:
hmm, @rabbah shed some light on the issue: "openwhisk operators can configure these limits at deployment time". Makes me wonder if more than 512MB should be allowed here. Right now the setting is just ignored. The warning is only shown when deploying with -v (verbose) and then there's so much output that it's very hard to notice...

Could we instead show a warning and still use the setting? What would this mean if the setting is actually invalid for the provider?

rabbah:
I haven’t looked at the implementation here but any hard coding of limits in the wskdeploy client is a bug in my opinion and should be read from the api host which provides them. 

There is an issue I opened related to the runtime kinds for example which bit me sometime ago.

mrutkows:
According to my recollection, wskdeploy itself imposes no limits on memory (or other size); thanks for referencing the docs/other issue.

rabbah:
Why is this closed?

TPei:
Just because the PR was merged, bu the underlying issue (hard coding client-side vs reading from api host) still remains, I guess. Updated the title accordingly

TPei:
the API (such as https://openwhisk.ng.bluemix.net/) returns the memory size limit as "limits.max_action_memory", this can be used accordingly I'd say

PR #1034, opened 2019-02-18
raise memorySize limit to 2048MB

TPei:
memorySize is limited to 512MB in wskdeploy, but not in OpenWhisk
#1033

### DIFF
--- a/utils/validation.go
+++ b/utils/validation.go
@@ -153,7 +153,7 @@ func LimitsMemoryValidation(memory *int) bool {
 	if memory == nil {
 		return true
 	}
-	if *memory < 128 || *memory > 512 {
+	if *memory < 128 || *memory > 2048 {
 		wskprint.PrintlnOpenWhiskWarning(wski18n.T(wski18n.ID_WARN_LIMITS_MEMORY_SIZE))
 		return false
 	}


TPei:
Ah, hang on, general OpenWhisk memory limit still seems to be 512MB:https://github.com/apache/incubator-openwhisk/blob/master/docs/reference.md

TPei:
I think the test failures are actually related to: https://github.com/apache/incubator-openwhisk/pull/4254, with things like `Invalid or missing runtime [swift:3.1.1] specified in manifest for the action [helloSwiftWithCode]`

utils/validation.go
```diff
@@ -153,7 +153,7 @@ func LimitsMemoryValidation(memory *int) bool {
 	if memory == nil {
 		return true
 	}
-	if *memory < 128 || *memory > 512 {
+	if *memory < 128 || *memory > 2048 {
 		wskprint.PrintlnOpenWhiskWarning(wski18n.T(wski18n.ID_WARN_LIMITS_MEMORY_SIZE))
 		return false
```
TPei: alternatively we could keep the warning for >512MB, but still return true

utils/validation.go
```diff
@@ -153,7 +153,7 @@ func LimitsMemoryValidation(memory *int) bool {
 	if memory == nil {
 		return true
 	}
-	if *memory < 128 || *memory > 512 {
+	if *memory < 128 || *memory > 2048 {
```
csantanapr: This is a quick fix for now, but I think the validation should be remove or do a validation against backend

csantanapr:
After this is merge the wsk CLI repo need to update the dependency file to pick up this bugs

rabbah:
I agree with Carlos. The limits should not be hardcoded in wskdeploy. This will likely need some systemic cleaning.

utils/validation.go
mrutkows: agree

Issue #1042, opened 2019-03-11
Can't specify limits on Sequences


TPei:
When settings limits on a sequence, such as
```yaml
packages:
  sequence_demo:
    actions:
      my_action:
        function: main.js
        limits:
          memorySize: 1024
    sequences:
      my_sequence:
        actions: >
          my_action,
          my_action
        limits:
          memorySize: 1024
```
it causes an error:
```
Error: manifest_parser.go [115]: [ERROR_YAML_PARSER_ERROR]: File: [manifest.yml]: 
==> yaml: unmarshal errors:
==> line 37: field limits not found in struct parsers.Sequence
```

However, Sequence memory limits can be set via the regular cli:
`wsk action update my_sequence --memory 1024`

PR #1043, opened 2019-03-11
allow limits on sequences

TPei:
Limits (like memorySize and timeout) should be allowed on sequences, since they can be set via the wsk cli
see #1042

### DIFF
--- a/parsers/manifest_parser.go
+++ b/parsers/manifest_parser.go
@@ -503,6 +503,13 @@ func (dm *YAMLParser) ComposeSequences(namespace string, sequences map[string]Se
 			wskaction.Annotations = annotations
 		}
 
+		// Limits
+		if sequence.Limits != nil {
+			if wsklimits := dm.composeActionLimits(*(sequence.Limits)); wsklimits != nil {
+				wskaction.Limits = wsklimits
+			}
+		}
+
 		// appending managed annotations if its a managed deployment
 		if utils.Flags.Managed || utils.Flags.Sync {
 			wskaction.Annotations = append(wskaction.Annotations, managedAnnotations)
--- a/parsers/manifest_parser_test.go
+++ b/parsers/manifest_parser_test.go
@@ -1499,9 +1499,9 @@ func TestComposeSequences(t *testing.T) {
 		switch wsk_action.Name {
 		case "sequence1":
 			assert.Equal(t, "sequence", wsk_action.Exec.Kind, "Failed to set sequence exec kind")
-			assert.Equal(t, 1024, wsk_action.Limits.Memory, "Failed to get Memory")
-			assert.Equal(t, 180, wsk_action.Limits.Timeout, "Failed to get Timeout")
-			assert.Equal(t, 1, wsk_action.Limits.Logsize, "Failed to get Logsize")
+			assert.Equal(t, 1024, strconv.FormatInt(int64(wsk_action.Limits.Memory.(int)), "Failed to get Memory")
+			assert.Equal(t, 180, strconv.FormatInt(int64(wsk_action.Limits.Timeout.(int)), "Failed to get Timeout")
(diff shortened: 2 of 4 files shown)


TPei:
hmmm :thinking: 
```
--- FAIL: TestComposeSequences (0.00s)
	manifest_parser_test.go:1502: 
			Error Trace:	manifest_parser_test.go:1502
			Error:      	Not equal: 
			            	expected: int(1024)
			            	actual  : *int((*int)(0xc420466968))
			Test:       	TestComposeSequences
			Messages:   	Failed to get Memory
	manifest_parser_test.go:1503: 
			Error Trace:	manifest_parser_test.go:1503
			Error:      	Not equal: 
			            	expected: int(180)
			            	actual  : *int((*int)(0xc420466978))
			Test:       	TestComposeSequences
			Messages:   	Failed to get Timeout
	manifest_parser_test.go:1504: 
			Error Trace:	manifest_parser_test.go:1504
			Error:      	Not equal: 
			            	expected: int(1)
			            	actual  : *int((*int)(0xc420466988))
			Test:       	TestComposeSequences
			Messages:   	Failed to get Logsize
```

pritidesai:
@TPei can you please try replacing `strconv.FormatInt(int64(wsk_action.Limits.Memory.(int))` ?

blt9001:
Is there anything I can do to help expedite this merge?

mdeuser:
while memory, logging and time limits _can_ be configured for sequences, i don't believe they actually have any effect.  do they?

blt9001:
@mdeuser I have asked this question as well - it is configurable via the cli - so I assumed it would have an effect, although I have done no formal testing.   When a sequence is created, it is given a default memory limit of 256 - and is configurable in the ibm cloud ui.   If it has no effect - that would be nice to clarify.  Otherwise - the change proposed would seem consistent with the feature/functions available through the cli and displayed in the ibm cloud functions ui.

TPei:
> @TPei can you please try replacing `strconv.FormatInt(int64(wsk_action.Limits.Memory.(int))` ?

Can you clarify what you mean by that @pritidesai ? Like this https://github.com/apache/openwhisk-wskdeploy/pull/1043/commits/75fea16ed1198e0790a192624dcb847320fbcc81 ?

HunderlineK:
I also have a question similar to @mdeuser, specifically for timeout though: is it the total time spent on all of the internal functions? e.g. if one step of the sequence is 100s and another 100s, but the limit of the sequence itself is 150s, will it fail before finishing the second step? Will the second step be killed?

mdeuser:
@HunderlineK - while not openwhisk doc, ibm functions doc https://cloud.ibm.com/docs/openwhisk?topic=cloud-functions-actions#actions_seq says
```
A sequence does not have limits that are separate from those of each action contained within the sequence, so although you can set limits for timeout, memory, and logsize when you create a sequence, those limits are ignored. Note that the limits configured for each action in the sequence are individually enforced. Because a sequence is a pipeline of actions, a failure in one action breaks the pipeline. For example, if one action times out, the entire sequence is exited with that failure.
```

TPei:
Ah, interesting @mdeuser, I guess than this whole thing is pointless. Can anyone confirm that this does indeed not make sense? @rabbah maybe?
I was just always confused that I had sequence actions with 128MB memory limit shown and thought that was annoying /shrug

rabbah:
It's correct that the limits are ignored/have no effect currently on a sequence. See also this long standing issue https://github.com/apache/openwhisk/issues/2396 meant to be a discussion to whether we wanted to change the behavior - for example, we could dictate that the sequence limits apply to every action and override individual settings. Alternatively, we can remove the limits entirely on a sequence and not allow them to be set/retrieved. In either case, one needs support in the backend to change the current semantics.

TPei:
cool, thanks for clarifying :)

Issue #1000, opened 2018-10-29
Client from "latest" binaries reports version is "latest"


afrittoli:
I just installed wskdeploy from the latest binary package for Mac amd64 (https://github.com/apache/incubator-openwhisk-wskdeploy/releases/download/latest/openwhisk_wskdeploy-latest-mac-amd64.zip). 

When asking for the version the client reports "latest" which is not very useful, since this information is obtained from the local binary, which means it's almost always stale.

```
$ wskdeploy version
wskdeploy version: latest
```

houshengbo:
@afrittoli Each commit to the master branch will kick-off the travis job to build the binaries with the tag "latest" automatically. The key work "latest" is used as the version, and we do not pick up any other key word for the version purpose.

A meaningful version name is added when a commit is tagged, so that the tag name is used as the version to generate the binaries.

AnthonyAmanse:
@houshengbo I think adding like a date and the latest commit SHA fingerprint would be useful. Something like `latest-12-05-2018-<commit-sha>`. This makes it easier for users to sync/compare their local `wskdeploy` binaries to what's on the upstream's "latest".

Side question: Is the `0.9.9` version the stable and recommended version?

rabbah:
Great suggestion @AnthonyAmanse  - are you interested in submitting a PR?

PR #1047, opened 2019-03-27
Added date and commit SHA to the version command

AnthonyAmanse:
This commit adds in a date and the head commit SHA. This makes it easy
for end-users to verify if their binary is indeed the latest one. The
date and git commit are also added in builds for release candidates.

Example:
```
./wskdeploy version
wskdeploy version: latest
wskdeploy git commit: 50aacbc7992720290f8a4a8696d2288d3c52a0e7
wskdeploy build date: 2019-03-28T20:41:05Z
```

Fixes: #1000 

Signed-off-by: AnthonyAmanse <ghieamanse@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ before_deploy:
       export GIT_TAG=$TRAVIS_TAG;
       export TAG=true;
   fi
-- ./gradlew --console=plain releaseBinaries -PpackageVersion=$GIT_TAG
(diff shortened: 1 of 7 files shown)


pritidesai:
yup I agree with @mrutkows, I like the date but commit hash is too granular, how does this impact on the released binaries? 

@houshengbo please weigh in your comments here.

mrutkows:
@pritidesai @houshengbo Perhaps the most reasonable plan to merge this and if anyone complains we can redact the display <or> add extra logic to display on when --verbose is also provided.  What do you think?

houshengbo:
This PR looks good to me. Adding git commit hash and the date makes it more self-elaborated for a build.

Issue #548, opened 2017-09-23
Description field/key is not supported for Rule entity


mrutkows:
In theory, Rules should support the "Shared Entity schema" defined in the spec. which includes a descriotion, displayName and Annotations.  However, the grammar in v0.8.9 clearly has a description field (we can discuss whether Annotations are supported on Rules; they may not be now but Rules should have more control logic and importance going forward in the programming model).

yamlparser.go struct:
type Rule struct {
	//mapping to wsk.Rule.Trigger
	Trigger string `yaml:"trigger"` //used in manifest.yaml
	//mapping to wsk.Rule.Action
	Action string `yaml:"action"` //used in manifest.yaml
	Rule   string `yaml:"rule"`   //used in manifest.yaml
	//mapping to wsk.Rule.Name
	Name string
}

pritidesai:
Annotations are supported in Rules.

alvarolopez:
This but is valid also for the Packages and Actions:

    Error: manifest_parser.go [115]: [ERROR_YAML_PARSER_ERROR]: File: [manifest.yaml]: 
    ==> yaml: unmarshal errors:
    ==> line 3: field description not found in struct parsers.Package
    ==> line 10: field description not found in struct parsers.Action

PR #1057, opened 2019-06-20
Support Description field in corresponding entities

alvarolopez:
The Description field should be supported across all entities, as it is
included in the "Shared Entity Schema".

Fixes: #548

### DIFF
--- a/parsers/yamlparser.go
+++ b/parsers/yamlparser.go
@@ -113,6 +113,7 @@ type Action struct {
 	Limits      *Limits                `yaml:"limits"`
 	Inputs      map[string]Parameter   `yaml:"inputs"`
 	Outputs     map[string]Parameter   `yaml:"outputs"`
+	Description string                 `yaml:"description,omitempty"`
 	Annotations map[string]interface{} `yaml:"annotations,omitempty"`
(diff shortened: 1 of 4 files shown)


mrutkows:
@alvarolopez I believe we held off placing support for it in the client since the Description is not yet recognized or stored at the server.  This would have to be done by using the object's metadata field (to not alter server schema) and corresponding changes to the Client / CLI to allow them to be read/written as well.

alvarolopez:
@mrutkows I believe that adding the description field worked for me, I need to check and keep you posted, as I remember some fields not working as you say.

mrutkows:
@alvarolopez I will merge this change allowing the "Description" to be part of the schema, but to be clear, any description data will not be respected at the server (i.e., not stored directly or in Annotation metadata) without further changes in wskdeploy to read/write this info. to the Annotations (OR get the server-side schema to allow Description to be a key on the actual Rule, Trigger, etc. entities directly).

alvarolopez:
@mrutkows I forgot to reply to you: we're fine with not storing the description at server-side, but having it as part of the schema so that every action can include a description. Many thanks.

PR #1065, opened 2019-08-08
Remove 'incubat(ing, ion, or)' designation and disclaimer from all files

mrutkows:


.travis.yml
```diff
@@ -74,9 +74,9 @@ before_deploy:
       git config --global user.email "builds@travis-ci.com";
       git config --global user.name "Travis CI";
       git tag -d $GIT_TAG;
-      git push -q https://$API_KEY@github.com/apache/incubator-openwhisk-wskdeploy :refs/tags/$GIT_TAG;
+      git push -q https://$API_KEY@github.com/apache/openwhisk-wskdeploy :refs/tags/$GIT_TAG;
```
chetanmeh: Would the existing api keys (encrypted for old repo) would still work

mrutkows: The API key portion of the push is likely not tied to the actual URL of the repo; auth keys typically are generated from UUIDs that are associated with locators and not directly derived from hashing them.

PR #1068, opened 2019-08-21
Auto supply a dummy API GW token

jiangpengcheng:
Since the `APIGW_ACCESS_TOKEN` is an optinal parameter, and the action `apimgmt/createApi` do require this parameter no matter what the value is, I think it's better to provide a fake token just like the [openwhisk-cli](https://github.com/apache/openwhisk-cli) did, see [api.go](https://github.com/apache/openwhisk-cli/blob/07993fd22c5618ee153eade186369d7d6084d1b7/commands/api.go#L1013):

related to: https://github.com/apache/openwhisk-wskdeploy/issues/1037


parsers/manifest_parser.go
```diff
@@ -1148,12 +1148,12 @@ func (dm *YAMLParser) ComposeApiRecordsFromAllPackages(client *whisk.Config, man
...
-	// verify APIGW_ACCESS_TOKEN is set before composing APIs
-	// until this point, we dont know whether APIs are specified in manifest or not
+	// supply a dummy API GW token as it is optional
 	if pkg.Apis != nil && len(pkg.Apis) != 0 {
 		if len(client.ApigwAccessToken) == 0 {
-			return nil, nil, wskderrors.NewWhiskClientInvalidConfigError(
-				wski18n.ID_MSG_CONFIG_MISSING_APIGW_ACCESS_TOKEN)
+			warningString := wski18n.T(wski18n.ID_MSG_CONFIG_MISSING_APIGW_ACCESS_TOKEN)
+			wskprint.PrintOpenWhiskWarning(warningString)
+			client.ApigwAccessToken = "DUMMY TOKEN"
```
mrutkows: Could we make the string "DUMMY TOKEN" a package constant and also "wskprint" that we are using a default string with the value "DUMMY TOKEN" so it is clear utility is providing a default behavior (due to lack of explicit value)?

mrutkows:
@jiangpengcheng Thanks for finding and making this correction. If you make these simple  changes requested in the review comments, I plan to merge.

jiangpengcheng:
ok, I use a constant for the dummy token, and print a more clear message for users

PR #1078, opened 2019-11-07
Add the ability to specify api via swagger in manifest

pwplusnick:
Adds the ability for swagger file specifying the api to be
  referenced in the manifest under project.config. The config
  name was chosen to be consistent with whisk-cli -c/--config
  option.

Note: I'll be adding a few documentation updates at the end of the day

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,8 +1,16 @@
 {
 	"ImportPath": "github.com/apache/openwhisk-wskdeploy",
 	"GoVersion": "go1.9",
-	"GodepVersion": "v62",
+	"GodepVersion": "v80",
 	"Deps": [
+		{
+			"ImportPath": "github.com/apache/openwhisk-client-go/whisk",
+			"Rev": "ee5b8709787cd37201c42e38040e9709f6d1e9c8"
+		},
+		{
+			"ImportPath": "github.com/apache/openwhisk-client-go/wski18n",
+			"Rev": "ee5b8709787cd37201c42e38040e9709f6d1e9c8"
+		},
 		{
 			"ImportPath": "github.com/cloudfoundry/jibber_jabber",
 			"Rev": "bcc4c8345a21301bf47c032ff42dd1aae2fe3027"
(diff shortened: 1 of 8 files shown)


mrutkows:
@we should add as a follow-on PR the i18n error/warning messages/strings specific to swagger (config) document parsing/mgmt.

Godeps/Godeps.json
```diff
@@ -164,9 +164,13 @@
...
-        {
-            "ImportPath": "github.com/palantir/stacktrace",
-            "Rev": "78658fd2d1772b755720ed8c44367d11ee5380d6"
-        }
+    {
+      "ImportPath": "github.com/palantir/stacktrace",
+      "Rev": "78658fd2d1772b755720ed8c44367d11ee5380d6"
+    },
+    {
+      "ImportPath": "github.com/ghodss/yaml",
```
pritidesai: Hey @pwplusnick, not sure how `github.com/ghodss/yaml` is different from `gopkg.in/yaml.v2`

deployers/servicedeployer.go
```diff
@@ -808,10 +810,17 @@ func (deployer *ServiceDeployer) DeployRules() error {
 
 // Deploy Apis into OpenWhisk
 func (deployer *ServiceDeployer) DeployApis() error {
-	for _, api := range deployer.Deployment.Apis {
-		err := deployer.createApi(api)
-		if err != nil {
-			return err
+	var err error
+	// Only deploy either swagger or manifest defined api, but not both
+	// Swagger API takes precedence
```
pritidesai: please add caps NOTE saying the same, Swagger API take precedence over the apis defined in manifest YAML.

deployers/servicedeployer.go
```diff
@@ -808,10 +810,17 @@ func (deployer *ServiceDeployer) DeployRules() error {
...
+	if deployer.Deployment.SwaggerApi != nil && deployer.Deployment.SwaggerApiOptions != nil {
+		err = deployer.createSwaggerApi(deployer.Deployment.SwaggerApi)
+	} else {
```
pritidesai: missing check on err here

docs/wskdeploy_apigateway_open_api_spec.md
```diff
@@ -0,0 +1,177 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+# API Gateway
```
pritidesai: rename it to something more specific to API specifications?

docs/wskdeploy_apigateway_open_api_spec.md
```diff
@@ -0,0 +1,177 @@
...
+
+This example builds on the ["Hello World" Action](wskdeploy_action_helloworld.md#actions) example and ["Hello World" API](wdkdeploy_apigateway_helloworld.md) example by adding an Open API Specification to define an API that is invoked via an API call.
+
+It shows how to:
+- update the Action named ‘hello_world’ to expose it to the gateway.
+- specify the API's endpoint that will trigger the action via an Open API Specification.
+
+### Manifest file
+#### _Example: “Hello world” action with project config to point to the API spec_
+```yaml
```
pritidesai: Instead of hardcoding one more time here, can we bring the content from the file? docs/examples/manifest_hello_world_apigateway_open_api_spec.yaml

docs/wskdeploy_apigateway_open_api_spec.md
```diff
@@ -0,0 +1,177 @@
...
+      version: 1.0
+      license: Apache-2.0
+      actions:
+        hello_world:
+          function: src/hello.js
+          web-export: true
+```
+### Open API Specification
+#### _Example: "Hello world" open api specification_
+```json
```
pritidesai: Same as above, can we bring the content from the file? 
docs/examples/open_api_spec.json

docs/wskdeploy_apigateway_open_api_spec.md
```diff
@@ -0,0 +1,177 @@
...
+  }
+}
+```
+*NOTE*: Some providers such as IBM may have changed these keys. For example, IBM uses `x-ibm-configuration` instead of `x-gateway-configuration`.
+
+There are two major differences from _"Hello World" API_ example:
+- the root key is now project as the open api specification is a project wide concept.
+- a new `config` key specifying where the Open API Specification is located.
+
+The `config` key under `project` in the manifest file specifies where the Open API Specification is located. The keyword `config` was chosen to remain consistent with the OpenWhisk CLI flag option. The Open API Specification describes in a JSON document the the base path, endpoint, HTTP verb, and other details describing the API. For example, the document above describes a GET endpoint at `/hello/world` that recieves JSON as input and returns JSON as output.
```
pritidesai: is it `config` or `config-file` in CLI? it says `config-file` here https://github.com/apache/openwhisk-cli/blob/07993fd22c5618ee153eade186369d7d6084d1b7/commands/api.go#L174

pritidesai:
@pwplusnick are we checking if an action is a web action before creating an API using the specification? please confirm. `wskdeploy` checks if the action specified in manifest is a web action if not its created as web action if an api is requested for that action.

pwplusnick:
@pritidesai It does not check to see if the action is a web action. In fact, this does not do much checking on the swagger at all. It does a few very basic checks to make sure the base bath is valid, but apart from that it does not check the swagger on the client side. It basically shoves the swagger up to the api gateway and looks for errors there.

docs/wskdeploy_apigateway_open_api_spec.md
pwplusnick: There is no way to do this in standard markdown unfortunately.

Godeps/Godeps.json
pwplusnick: I can rewrite it, but it is used in a function I lifted from openwhisk-cli to do the swagger parsing. Ideally, we move that function to openwhisk-client-go and that dependency will get removed.

PR #1083, opened 2020-01-08
Enable setting user-supplied auth tokens on API create via require-whisk-auth annotation

mrutkows:


mrutkows:
Fixes https://github.com/apache/openwhisk-wskdeploy/issues/1046

PR #1092, opened 2020-03-24
Remove misleading web-export key

style95:
`web-export(key)` is being interchangeably used with `web` in `wskdeploy`.
On the other hand, it is differently used in `wsk` cli.

For example, if the `web` flag is enabled OW will make the action as a web action and enable the `final` flag. To avoid the use of protected parameters, users can configure two annotations, `web-export(true)` and `final(false)` flag.

But with `wskdeploy`, `web-export` is also being used as a top key of actions and it enables the `final` flag all the time.

Even if users still can control this with annotations, I think it is misleading to have the same option in two different places with two different meanings.

So I removed the top-level one.

### DIFF
--- a/docs/examples/manifest_hello_world_apigateway.yaml
+++ b/docs/examples/manifest_hello_world_apigateway.yaml
@@ -23,7 +23,8 @@ packages:
     actions:
       hello_world:
         function: src/hello.js
-        web-export: true
+        annotations:
+          web-export: true
     apis:
       hello-world:
         hello:
--- a/docs/examples/manifest_hello_world_apigateway_http.yaml
+++ b/docs/examples/manifest_hello_world_apigateway_http.yaml
@@ -23,10 +23,12 @@ packages:
     actions:
       hello_world:
         function: src/hello_http.js
-        web-export: true
+        annotations:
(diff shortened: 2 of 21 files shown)


mrutkows:
There is historical context for this web-export being added at a higher level than annotations... that is, we had intended that wskdeploy manifests would not allow/expose annotations period (at all).  However, we were forced by some use cases to add support for it.

My concern is that the code SHOULD still permit the web-export key outside annotations (that is still work and not break anyone using it for 3 years), but deprecate it.  Then update all examples use cases to use the annotations, but preserve at least one functional test to make sure that, although deprecated, it still works.

style95:
@mrutkows Thank you for the good comment.
I would update it accordingly.

style95:
@mrutkows 
I initially thought of explicitly specifying the deprecation of the top-level web-export field.
But since both top-level web-export and `annotations.web-export` can coexist, I think it would be enough to just change the top-level one to the annotations' one in all occurrences.
Newbies will start using the `annotation.web-export` as the top-level one is no more exposed.

Do you think we have to explicitly specify the deprecation?

PR #1094, opened 2020-04-20
fix: add spec_parameter_types.md link for `type` property

shazron:
This link is to clarify what constitutes valid values for the `type` property for `parameter`s.


shazron:
Not sure why the build errored, this is just a doc change. The build was running for 6.5 hrs so it might be a Travis thing. However, I don't have the karma to re-start the build.

mrutkows:
@shazron Thanks for suggesting this hyperlink; I added it as part of a larger docs update as part of PR https://github.com/apache/openwhisk-wskdeploy/pull/1120

PR #1098, opened 2020-08-04
Add dotnet core 3.1 action support

woutersterp:
I think this is the only thing to be added for dotnet core 3.1 support right?

### DIFF
--- a/runtimes/runtimes.go
+++ b/runtimes/runtimes.go
@@ -468,12 +468,27 @@ var RUNTIME_DETAILS = []byte(`{
         "dotnet": [
             {
                 "kind": "dotnet:2.2",
-                "default": true,
+                "default": false,
                 "deprecated": false,
                 "requireMain": true,
                 "image": {
                     "prefix": "openwhisk",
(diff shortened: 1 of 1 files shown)


mrutkows:
Please rebase and add some "helloworld" style unit and integration tests (manifests) to verify support in our CI/CD process (Travis).  Please reference the work done for the updated Rust runtime in this PR to get an idea of what the balance of work is needed here: https://github.com/apache/openwhisk-wskdeploy/pull/1119

runtimes/runtimes.go
```diff
@@ -480,6 +480,21 @@ var RUNTIME_DETAILS = []byte(`{
                     "attachmentName": "codefile",
                     "attachmentType": "text/plain"
                 }
+            },	    
+            {
+                "kind": "dotnet:3.1",
```
mrutkows, line 485: Please add unit tests and perhaps an integration test to validate support in our CI/CD process. Thanks!

mrutkows (changes requested):

PR #1101, opened 2020-08-10
Migrate to using Go Module to manage dependencies

huydoan2:
Besides using Go Module (adding go.mod and go.sum), there are several changes:

1. Upgraded Go version to 1.14 alongside with other packages.
2. Modified unit tests to accommodate changes.


mrutkows:
Appreciate your work on this Huy... hopefully, this a will address the long-standing desire to use go mod as referenced in this issue: https://github.com/apache/openwhisk-wskdeploy/issues/1082

tests/dat/manifest_data_compose_actions_for_limits.yaml
```diff
@@ -21,7 +21,6 @@ packages:
       hello1:
         function: ../src/integration/helloworld/actions/hello.js
         limits:
-          timeout: 1
           timeout: 600000
```
rabbah, line 24: Should this be 60000 ?

huydoan2: Not sure. I didn't change that value. Additionally, that test file is used in parsing tests so the values may not matter much.

mrutkows: The value was not intended to be 60 seconds... just testing input matches output against some larger (max) value. Here 10 minutes...

mrutkows:
@style95 Hi Dom, could you and colleagues (who recently showed wskdeploy IDE plugins) please review so we make sure we are not going to break you if we move to merge.  Thanks!

upgle:
Thank you for your contribution 🎉 

Could you please update the README.md? Some contents need to be modified for `go mod`.
https://github.com/apache/openwhisk-wskdeploy#developers-should-use-go-deps-and-go-build-not-go-get

And I've checked there is no issue for VSCode openwhisk plugin.
@KeonHee Could you check it for Jetbrains extension?

KeonHee:
@mrutkows @upgle Thanks for noti & @huydoan2 Thanks for your contribution :)

It works well with Jetbrain plugin too.

huydoan2:
Thank you, all.

@upgle We will need to clean up the doc and the Makefile in a separate PR :).

PR #1112, opened 2021-02-10
Generate hosts script

style95:
Generate the hosts file automatically.


rabbah (approved):

mrutkows:
@style95 Thanks! I agree we should use the hosts file from the core repo.

mrutkows (approved):

PR #1115, opened 2021-02-12
Remove unsupported Darwin-386 (23-bit) from GoLang 1.15 build matrix

mrutkows:
The wskdeploy (release, non-PR) builds including "latest" have been failing since upgrading to GoLang 1.15 since the 32-bit architecture is no longer supported for Darwin as a decision from the golang project.  

See "remove darwin 386" PR: https://github.com/golang/go/issues/37610

I am removing it from our gradle build matrix to hopefully get our binary release builds running again...


pritidesai (approved):
👍 thanks @mrutkows

PR #1118, opened 2021-02-18
Fix gogradle plugin to stop using vendor directory

mrutkows:
ideally, I would love to replace the `gogradle` plugin (no sig. updates since 2017) with something more aware of the post go 1.11 moduling system and supports cross platform/arch builds simply.  However, we need a means to prevent `gogradle` from using a "/vendor" directory and attempting to fetch its own packages/versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,6 @@ git clone --branch 1.1.0 git@github.com:apache/openwhisk-wskdeploy
 
 You can also pull the code from a fork of the repository. If you intend to become a Contributor to the project, read the section [Contributing to the project](#contributing-to-the-project) below on how to setup a fork.
 
-
 ### Building using `go build`
 
 Use the Go utility to build the ```wskdeploy``` binary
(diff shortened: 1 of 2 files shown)


mrutkows:
My assumption is that the gogradle plugin forces you to use a "/vendor" directory and we really do not want one at all; however, the "go" command assumes that if it sees a "/vendor" directory that it overrides "go.mod".  So we need to either prevent gogradle from installing anything into "/vendor" or we set the new flag on the "go" command i.e., `-mod=mod` to force it to ignore the vendor directory.  

See: https://tip.golang.org/doc/go1.14 for a desc. of this behavior

mrutkows:
I should mention that a 3rd option exists, but I discounted it since it basically capitulates to the old gogradle plugin (and also bypasses it for managing deps.) and uses the "go" command to create the vendor directory i.e., `go mod vendor`: https://www.honeybadger.io/blog/golang-go-package-management/

mrutkows:
ideally, we switch to a more current gradle plugin such as: https://github.com/innobead/gogradle which works better with go.mod and is more straightforward or eliminate gradle altogether and find a better way to switch OS/ARCH for docker builds.

pritidesai (approved):
thanks @mrutkows for the detailed explanation, hope this fixes the build 🤞

mrutkows:
@pritidesai Thanks Priti!

PR #1119, opened 2021-02-20
Rust 1.34 support 

kingledion:
Adding support for kind "rust:1.34" in wskdeploy. 

This is tested working in local deployments with cargo and source files zipped, but is not working with plain '.rs' extension files.


mrutkows:
@kingledion Thanks for your contribution! It is quite a good coincidence as I had intended to figure out how to bring in the Rust tests from this old PR https://github.com/apache/openwhisk-wskdeploy/pull/1074. 

Effectively this PR:
Fixes https://github.com/apache/openwhisk-wskdeploy/pull/1074

If there are still mapping issues in the util. from seeing a `.rs` extension to the actual runtime; can we open an issue and track/debug/work on as a follow-on? Thanks!

mrutkows (approved):
Thanks for your work filling out the intrinsic support for the Rust runtime!

PR #1122, opened 2021-02-24
Streamline README for Gradle and cross-architecture builds & upgrade gogradle plugin

mrutkows:
The Gradle task usage instructions were really outdated and unnecessarily cumbersome, especially in regard to pre-reqs and for building for diff. platforms/architectures.  Really tried to simplify as well as distinguish between use of gradle wrapper and local gradle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ Packages/
 
 # Gradle
 .gradle
+.gogradle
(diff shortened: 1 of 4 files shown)


pritidesai (approved):
thanks @mrutkows

PR #1124, opened 2021-03-03
Update openwhisk-client-go to release 1.1.0 as well as other pkg dependencies

mrutkows:


mrutkows:
Please note that after this PR is merged the only pks that wskdeploy relies upon that are not based on a release (tag) are as the following:

- github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32 
>>> which is the latest commit available and includes its own upgrade to gomod, and

- github.com/cloudfoundry/jibber_jabber v0.0.0-20151120183258-bcc4c8345a21
>>> which has had not had any commits since Nov 2015 and never produced any releases or tags... ideally, we would look to replace this very small function used to detect current system language

pritidesai (approved):
wow, lot of cleanup, thanks @mrutkows

dgrove-oss (approved):

PR #1131, opened 2021-03-23
Update build.gradle

advancedwebdeveloper:


mrutkows:
Although I am curious to see if this builds, the direction of our Go projects/tools that use Gradle and the gogradle plugin is one where we integrate a newer/supported plugin for Go that does not assume (old/deprecated) go vendor or one where we seek to eliminate gradle or a need for a plugin altogether (as the "go" tools have matured and become quite simple as to not require plugins perhaps at all).

